PR #10, opened 2025-05-13
Add comments that will be ignored in the tree

WebReflection:
This MR allows developers to add comments within the template code without polluting the final output with those comments:

```html
<!--# this comment won't exist live #-->
<!-- but this one will -->
<!--# ... or this one -->
<!-- or this one ... #-->
```


pauleveritt:
I see you retained the wrapping `<!--#1--><!--3#-->`. That doesn't how HTML comments work, so I was surprised.

pauleveritt:
My comment was wrong, the `test_dev_comments` covers both cases, a valid start/stop and an invalid start/stop.

PR #13, opened 2025-05-13
Fix #12 - Allow raw templates as interpolated values

### LINKED ISSUE #12: Omit inner `html()` calls
### Discussed in https://github.com/t-strings/tdom/discussions/11

<div type='discussions-op-text'>

<sup>Originally posted by **pauleveritt** May 13, 2025</sup>
(I discussed this with @WebReflection in Discord.)

As an example, this fails:

```python
inner = t'<{Heading} title="My Title" />'
result = html(t"<div>Hello {inner}</div>")
```

The inner `Template` isn't evaluated:

```
'<div>Hello Template(strings=(&#x27;&lt;&#x27;, &#x27; title=&quot;My Title&quot; /&gt;&#x27;), interpolations=(Interpolation(&lt;function Heading at 0x103e96c90&gt;, &#x27;Heading&#x27;, None, &#x27;&#x27;),))</div>
```

When using deeply-nested component calls, the extra `html()` can be cumbersome. It would be nice to build up a tree that includes unrendered `Template` then, later in the chain, render to a string.</div>

WebReflection:
This MR allows the usage of raw templates as part of the interpolations that are resolved via either `html` or `svg`.

### DIFF
--- a/src/tdom/__init__.py
+++ b/src/tdom/__init__.py
@@ -15,6 +15,7 @@ try:
 except ImportError:
   pass
 
+
 def _util(svg):
   def fn(t):
     if not isinstance(t, Template):
@@ -22,7 +23,7 @@ def _util(svg):
 
     strings = t.strings
 
-    values = [entry.value for entry in t.interpolations]
+    values = [_resolve(fn, entry) for entry in t.interpolations]
 
     length = len(values)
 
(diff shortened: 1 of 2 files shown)


pauleveritt:
As @WebReflection commented in Discord, we need more test cases to think through what we want. For example, how much recursion, how this fits into middleware, etc.

WebReflection:
converting this as draft because I believe this is not what we want/need and it won't work with more than a single level

davepeck:
This is supported as part of the updated `tdom`

PR #20, opened 2025-06-18
Cleanup tests

pauleveritt:
Remove the old viewdom tests, fill in some examples and docs.


WebReflection (approved):

WebReflection:
@pauleveritt I'd like to publish latest tdom that doesn't break in MicroPython ... do you want these in too or do these matter?

pauleveritt:
@WebReflection No need to affect the release, as these changed aren't a part of the package.

PR #22, opened 2025-06-19
Remove delayed import

pauleveritt:
Remove the delayed import workaround for pytest-playwright since we aren't using it.

### DIFF
--- a/src/tdom/__init__.py
+++ b/src/tdom/__init__.py
@@ -6,14 +6,7 @@ from .dom import Node, Comment, DocumentType, Text, Element, Fragment, parse, un
 _parsed = {}
 _listeners = []
 
-# T = (t'').__class__
-# Let's just use the imported class
-try:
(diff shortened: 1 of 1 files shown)


src/tdom/__init__.py
```diff
@@ -6,14 +6,7 @@
...
 
-# T = (t'').__class__
-# Let's just use the imported class
-try:
-  # Temporary hack because pytest-playwright has to run under 3.13
-  # in CI due to greenlet not compiling there under 3.14.
-  from string.templatelib import Template
-except ImportError:
-  pass
+from string.templatelib import Template
```
WebReflection, line 9: does this work on MicroPython port? that's also the reason it was there

pauleveritt: If the MicroPython flavor supports t-strings, it works. If not, then tdom isn't very useful. :)

PR #24, opened 2025-08-15
MicroPython tests and an examples web app

pauleveritt:
Make development and exploring under MicroPython easier:

- Directly include a local MicroPython build that works fast/offline with t-strings support
- pytest strategy with Playwright and network interceptors to run without needing an HTTP server
- A web app `index.html` at the top which loads all the examples into an editable/runnable browser page

To Do:
- Switch from Polyscript to PyScript (pathlib dependency is blocker)
- Much better UI needed, load all the examples, etc.
- Consider switching to the PyScript editor widget
- Decide if we want the MicroPython binaries in the repo directly, or create a CLI script to go fetch them
- Same for Polyscript NPM checkout, bundled, or CLI
- Bring back GitHub Actions testing


pauleveritt:
I updated to use a local copy of PyScript. I will pause on this until finding out if dev/test/app for MicroPython is useful. Also, wait for @WebReflection to wrap up his tdom rethink.

Lots left to do if we proceed:
- Stop using CodeMirror 6 and switch to PyScript's CM5
- Nicer UI
- Browse between example sets and bring all the other examples in
- Address why every 5th or so reload gets the updates error

PR #39, opened 2025-09-12
Draft attempt at a context object

pauleveritt:
The `html()` call takes an option `context=` named argument. This is then passed down the calling tree. If a component wants the context, it can add as a positional argument in the signature.

Tests cover both function components and class components.

Also add a section in the README.


davepeck:
Closing for now; we'll revisit context objects soon!

PR #52, opened 2025-09-14
Support generator components and explicit fragment syntax (`"<><li>1</li><li><2></li></>"`)

### LINKED ISSUE #48: Components should be able to return a generator
This example:

```python
```python
from typing import Iterable

def Todos() -> Iterable[Template]:
    for todo in ["first", "second", "third"]:
        yield t"<li>{todo}</li>"

result = html(t"<ul><{Todos} /></ul>")
assert str(result) == '<ul><li>first</li><li>second</li><li>third</li></ul>'
```

...fails with this traceback:

```
Traceback (most recent call last):
  File "/Users/pauleveritt/Library/Application Support/JetBrains/PyCharm2025.2/scratches/scratch_8.py", line 12, in <module>
    result = html(t"<ul><{Todos} /></ul>")
  File "/Users/pauleveritt/projects/t-strings/tdom/tdom/processor.py", line 385, in html
    return _substitute_node(p_node, template.interpolations)
  File "/Users/pauleveritt/projects/t-strings/tdom/tdom/processor.py", line 362, in _substitute_node
    new_children = _substitute_and_flatten_children(children, interpolations)
  File "/Users/pauleveritt/projects/t-strings/tdom/tdom/processor.py", line 270, in _substitute_and_flatten_children
    substituted = _substitute_node(child, interpolations)
  File "/Users/pauleveritt/projects/t-strings/tdom/tdom/processor.py", line 364, in _substitute_node
    return _invoke_component(tag, new_attrs, new_children, interpolations)
  File "/Users/pauleveritt/projects/t-strings/tdom/tdom/processor.py", line 341, in _invoke_component
    raise TypeError(
    ...<2 lines>...
    )
TypeError: Component callable must return a Node, Template, or str; got generator
```

Suggested change in `processor._invoke_component`:

```python
        case Iterable():
            # Support generator/iterable returns from component callables
            children: list[Node] = []
            for item in result:
                # Each yielded item can be a Node, Template, str, HasHTMLDunder
                match item:
                    case Node():
                        children.append(item)
                    case Template():
                        children.append(html(item))
                    case str():
                        children.append(Text(item))
                    case HasHTMLDunder():
                        children.append(Text(Markup(item.__html__())))
                    case _:
                        # Fallback to string conversion to mirror _node_from_value behavior
                        children.append(Text(str(item)))
            return Fragment(children=children)
```

davepeck:
When merged, resolves #48


davepeck:
> Perhaps I'm misunderstanding that README line?

@pauleveritt 

Ah. It means you can return an `Element` instance; that is what `html(t"<div>Hello</div>")` will produce and what `Hello()` does indeed return.

It doesn't mean you can set the return type _annotation_ to `Element` since `html()` is defined as returning `Node`, of which `Element` is only one subclass. Pyright can't be sure of the type you've gotten back.

Maybe we can rewrite this for clarity.

pauleveritt:
Ok, got it. I can update the README and fix that usage.

Let's say I'm writing a test and it wants to do `result.props`. That's on `Element` not `Node`. Would I do a cast?

davepeck:
> Let's say I'm writing a test and it wants to do `result.props`. That's on `Element` not `Node`. Would I do a cast?

Yes, a `typing.cast(Element, html(t"<div>Hello</div>"))` would work. But in the context of a test, I'd probably do an assert:

```python
node = html(t"<div>Hello</div>")
assert isinstance(node, Element)
# Now you can safely look at node.children
```

Issue #30, opened 2025-09-10
Consider supporting substitution in mid-attribute


davepeck:
For example:

```python
result = html(t'<hr data-something="foo {42} bar" />')
```

We don't currently support this and I'm not sure there's any sensible performant way _to_ support this. Also, it's not clear what types of values we would support in practice.

But we should consider it!

Alternatively, we should rule this out, document that fact, and figure out how to raise an exception if it happens accidentally.

PR #59, opened 2025-10-01
Enable string interpolation with multiple placeholders in HTML attribute

koxudaxi:
This pull request enhances the handling of interpolated attribute values in the `tdom` processor, allowing for embedded and multiple placeholders within string attributes. The main improvement is the ability to substitute placeholders that are not the entire attribute value, supporting more flexible and expressive template syntax. The changes are covered by new tests to ensure correct behavior.

**Improvements to attribute interpolation:**

* Added `_PLACEHOLDER_PATTERN` using regular expressions to match placeholder patterns within strings, enabling detection of embedded placeholders. (`tdom/processor.py`) [[1]](diffhunk://#diff-2dc811a3da78c4795d089a26c8adb4254a10aeb2c677c0081aa97d5f1f244fafR2) [[2]](diffhunk://#diff-2dc811a3da78c4795d089a26c8adb4254a10aeb2c677c0081aa97d5f1f244fafR57)
* Introduced `_replace_placeholders_in_string` to replace any number of placeholders within a string attribute value, handling both static and dynamic segments. (`tdom/processor.py`)
* Updated `_substitute_interpolated_attrs` to use the new pattern and replacement logic, supporting attributes with embedded and multiple placeholders, not just those consisting solely of a placeholder. (`tdom/processor.py`)

**Testing and validation:**

* Added tests for embedded, prefixed/suffixed, and multiple placeholders in attribute values to verify new functionality. (`tdom/processor_test.py`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,15 @@ button = html(t"<button id={element_id}>Click me</button>")
 # <button id="my-button">Click me</button>
 ```
 
+Multiple substitutions in a single attribute are supported too:
+
+```python
+first = "Alice"
+last = "Smith"
+button = html(t'<button data-name="{first} {last}">Click me</button>')
+# <button data-name="Alice Smith">Click me</button>
+```
+
 Boolean attributes are supported too. Just use a boolean value in the attribute
 position:
 
--- a/tdom/processor.py
+++ b/tdom/processor.py
@@ -103,7 +103,7 @@ def _replace_placeholders_in_string(
         return dynamic_segments[0][1]
 
     return "".join(
-        segment[1] if segment[0] == "static" else str(segment[1])
+        str(segment[1]) if segment[0] == "static" else str(segment[1])
         for segment in segments
     )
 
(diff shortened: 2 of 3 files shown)


davepeck:
@koxudaxi This is great, thanks! And it resolves issue #30 entirely.

I added a few tests to document (and make sure I understood!) the behavior for "special" attributes like `data`, `aria`, `style`, and `class`.

I think I like it because it boils down to a simple rule: if you have multiple placeholders in an attribute value, that value is treated as a string. For `data` and `aria`, the value _must_ be `dict`-like, so we raise an exception. For `style` and `class`: you get a string; those attributes _do_ accept strings, so it's fine.

WebReflection:
late to the party, but this is a slippery slope and neither React nor native DOM on the Web supports this ,,, happy you are happy but thinking that an attribute as a whole can react to inner changes is misleading, while `atr={f"a b {x} {y}"}` is honest.

PR #65, opened 2025-10-15
Add missing sections in README.md.

sfermigier:
Also, run tests on the README.


davepeck (approved):
Oh wow, @sfermigier -- thank you very much!

Some of the documentation is likely to change (maybe significantly, for instance about context), but this is a great capturing of the current pre-alpha state of the library.

Very very much appreciated.

Issue #68, opened 2025-10-27
Not escaping "content" elements might be a security issue


ianjosephwilson:
Seems that it would be possible for someone to interpolate in a "content" element and not realize that values are not escaped.  This seems like a grey area but opt-in is more likely to be expected?  I'm not sure what would fix this but I think the `__str__` method of `Element` might be the issue.  I just tried to make a test to demonstrate the issue.

An example I'll try to create a test for might be to inject a closing tag into a style tag, then a script tag and then re-start the style as if nothing happened.

WebReflection:
sure thing if you have `<script>{user_input}</script>` you have more questions to ask yourself than anything this template string idea could solve or answer

WebReflection:
TL;DR (and imho) ... you are focusing on the wrong thing in here ... the current behavior is specs compliant, it allows a server owning its template string to even append/produce `t"</script><script src="{polyfill_or_solution_or_special_case}"></script>"` when desired, sanitizing stuff that should not need sanitization would make this template potentials less appealing for real-world use cases ... if people out there expect `<script>{thing}</script>` to be automatically augmented in terms of intent in a way that satisfies both Server and Front End, there is an issue around safety and what user inputs mean VS server side developers intents ... and that's both a mistake and a slippery slope if not already a killing potentials for the templating engine based on t-strings VS everything else.

**edit** wanna put anything into a `<textarea>` that was not meant by the server? `t'<textarea>{sanitize(input)}</textarea>` it is, anything else is a mistake to me.

WebReflection:
to clarify:

  * remove `is_content` logic and you get partial `<script>what = {ever}</script>` out of the box ... that's not possible for FE facing libraries unless tons of extra logic is in place because `script` is a *data* element
  * state this library is only for Python server side templating because otherwise it would break on the FE side even with Pyodide or other C-Python based alternatives because in there real DOM is created and real DOM has its own rules

WebReflection:
after thoughts: you'd be better off with xml parsing that produces at the end html instead ... that would remove a lot of "friction" around what's valid or not as content for special elements, as everything will be easy-peasy out of the box ... although you lose the **void** elements ability to never get closed, such as `<br>` or `<hr>` but you could sanitize that upfront to produce `<br />` or `<hr />` so that the outcome will be compatible with HTML 5+

**edit**

[...]

PR #70, opened 2025-10-29
Prospective fix for #68

davepeck:
Potential fix for #68.

### DIFF
--- a/tdom/nodes_test.py
+++ b/tdom/nodes_test.py
@@ -1,4 +1,5 @@
 import pytest
+from markupsafe import Markup
 
 from .nodes import Comment, DocumentType, Element, Fragment, Text
 
@@ -45,6 +46,14 @@ def test_text_safe():
 
     text = Text(CustomHTML())
     assert str(text) == "<b>Bold Text</b>"
+
+
+def test_text_equality():
+    text1 = Text("<Hello>")
+    text2 = Text(Markup("&lt;Hello&gt;"))
+    text3 = Text(Markup("<Hello>"))
+    assert text1 == text2
+    assert text1 != text3
 
 
 def test_fragment_empty():


tdom/parser.py
```diff
@@ -59,7 +61,7 @@ def handle_endtag(self, tag: str) -> None:
         self.append_element_child(element)
 
     def handle_data(self, data: str) -> None:
-        text = Text(data)
+        text = Text(Markup(data))
```
ianjosephwilson: You know more about this than me but it seems that Text() is running double, triple or even quadruple duty. Something about all-clearing this across the board seems suspect. (I copied this comment from commit)

davepeck: This is safe, in the sense that `Text` created when parsing HTML content at this layer of the stack is always trusted.

But I fully agree this highlights a potential confusion of responsibilities that needs to be teased apart. Currently, `Text.str` is treated as untrusted by default, unless explicitly marked otherwise (via `Markup` or compatible mechanism).

ianjosephwilson:
I think removing the edge-case check from rendering, in Element.__str__, is a big security improvement for a common exploit.

I'm not so sure about whitelisting the data coming out of `handle_data`.  I think maybe entities are not escaped correctly?  

```
________________________________________________________________________________ test_entities_are_reescaped _________________________________________________________________________________

    def test_entities_are_reescaped():
        node = html(t"<p>&lt;/p&gt;</p>")
        assert node == Element(
            "p",
            children=[Text("</p>")],
        )
>       assert str(node) == ("<p>&lt;/p&gt;</p>")
E       AssertionError: assert '<p></p></p>' == '<p>&lt;/p&gt;</p>'
E         
E         - <p>&lt;/p&gt;</p>
E         + <p></p></p>

tdom/processor_test.py:174: AssertionError

```

davepeck:
> I want to sit on it for at least a little while, to mull over whether this whole thing is a house of cards that needs to be reassembled or whether we're on the right track.

I'm comfortable with these changes. There really is an edge case for content nodes, but it was previously handled incorrectly. I think this set of changes pushes the content node handling to the right layer. 

@ianjosephwilson LMK if you still have pushback. Thanks again!

PR #72, opened 2025-11-06
Use markers

ianjosephwilson:
This is less invasive than my discussion.  It permits overriding a static attribute with a dynamic attribute either with a new value or to unset/remove it.  I added `data-*` tests but the other special cases need more tests.  Let me know what you think.

### DIFF
--- a/tdom/processor.py
+++ b/tdom/processor.py
@@ -211,13 +211,7 @@ def _instrument_and_parse(template: Template) -> Node:
 # Placeholder Substitution
 # --------------------------------------------------------------------------
 class AttrMarker(Enum):
-    """
-    Use attr markers to keep track of intention AND actual value during
-    processing.
-    """
-    REMOVE_ATTR_FALSE = auto()
-    REMOVE_ATTR_NONE = auto()
-    BARE_ATTR_TRUE = auto()
+    # Use this to mark attributes parsed in that are bare/boolean true.
     BARE_ATTR_NONE = auto()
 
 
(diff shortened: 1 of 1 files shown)


tdom/processor_test.py
```diff
@@ -621,6 +621,36 @@ def test_interpolated_data_attributes():
...
+    assert node == Element('div', {"data-selected": None})
+    assert str(node) == '<div data-selected></div>'
+
+
+def test_data_attr_unrelated_unaffected():
+    node = html(t'<div data-selected data={dict(active=True)}></div>')
+    assert node == Element('div', {"data-selected": None, "data-active": None})
+    assert str(node) == '<div data-selected data-active></div>'
+
+
```
davepeck: These tests are good additions...

tdom/processor_test.py
```diff
@@ -662,6 +692,20 @@ def test_interpolated_style_attribute():
...
 
+def test_clear_static_style():
+    node = html(t'<p style="font-color: red" {dict(style=None)}></p>')
+    assert node == Element("p")
+    assert str(node) == "<p></p>"
+
+def test_override_static_style_str_str():
+    node = html(t'<p style="font-color: red" {dict(style="font-size: 15px")}></p>')
+    assert node == Element("p", {"style": "font-size: 15px"})
+    assert str(node) == '<p style="font-size: 15px"></p>'
```
davepeck: This gets down to whether we want to support special-case merging for `style` and others, aka the `reduce()` approach. But, reading this test, perhaps it just makes sense to punt on that for now... as-is, this is simple and straightforward.

davepeck (commented):
Overall this is good. I do want to think more carefully about the return types from `_process_*_attr(...)` here...

tdom/processor_test.py
ianjosephwilson: Yeah you can imagine how `None` might cause confusion here.  I kind of think `Element` should be storing the attributes as their normal values and they are not converted to "parser/serialized" values until the actual html strings are generated.

PR #75, opened 2025-11-20
Clarify attribute interpolation pipeline

ianjosephwilson:
Another attempt at normalizing how attributes are resolved in the pipeline.
- Attempts to prevent type coercion of component attributes when using spread attributes.
- Attempts to prevent filtering out of `True` values that are prematurely coerced to `None` and then accidentally filtered out by misinterpreting `None` in a second call to `_process_attr`.
- Coerces bare attributes to `True` within the pipeline
    - Exports back to `None` before going to html
    - Exports as `True` when passing to components
- Does not apply "special attribute" handling to static attributes, ie. those that involve no interpolations.
- This (as before) does relax the type signature of `_process_data_attr` to `t.Iterable[tuple[str, object | None]]`.  I think as things go forward and the "special" attribute handling is adjusted and the interaction between different interpolations bringing in and out attributes is hardened it will make sense for all the `process_*_attr` functions to be dumping everything in as `[str, object|None]` with the intention for it to go straight to a component or get serialized or re-serialized back to "html" typed attributes, ie. `dict[str, str|None]` at the LAST stage.

Let me know what you think when you get a chance.  Thanks!

### DIFF
--- a/tdom/processor.py
+++ b/tdom/processor.py
@@ -308,6 +308,19 @@ def _process_attr(
     yield (key, value)
 
 
+def _process_static_attr(
+    key: str, value: str | None
+) -> t.Iterable[tuple[str, object | None]]:
+    """
+    Bring static attributes, parsed from html but without interpolations, into the pipeline.
+    """
+    match value:
+        case None:
+            yield (key, True)
+        case _:
+            yield (key, value)
+
+
 def _substitute_interpolated_attrs(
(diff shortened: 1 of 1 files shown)


davepeck:
@ianjosephwilson This looks good to me! Are you okay if I squash/merge, or are you still considering further work here?

ianjosephwilson:
Awesome, please merge when you get a chance.

davepeck:
Done and done! Thanks @ianjosephwilson !

PR #76, opened 2025-12-02
Potential structured/templated node design

ianjosephwilson:
I mentioned this concept in a discussion but had to clean it up a bit.  I wanted to get your thoughts about this approach.  I could keep cleaning this up or this could just be used as a roadmap.

#### Potential changes

- Added "template" nodes: `TNode`, `TElement`, `TFragment`, `TComment`, `TText`
    - The idea being that they are `Node` templates that are meant to be interpolated to create the actual `Node`s.
    - Store templated text in a `Template` object on `TComment` and `TText` (The "template" equivalents of `Comment` and `Text`).
    - Store templated attributes in a consolidated sequence of `str | int` on `TElement`.
    - A `TElement` is just informally considered a component right now based on if `component_info` is set or not.
    - `Node` inherits from `TNode` (SEE: node-inherits-from-tnode)
- Update `parser.py` to remove the placeholders
    - Externally the parser now considers an entire `Template` rather than being gradually fed chunks.
    - Placeholders are inserted as before but are removed when found during parsing:
        - component start/end tags
        - attributes
        - within an element's text
        - within a comment's text
    - If a placeholder is not found or found in a weird location an error is raised.
    - Parser now implements `handle_startendtag` (SEE: special-endtag-handling-in-parser)
    - The "subtemplate" coordinates(?) of a component are tracked within the larger template (SEE: templates-embedded-in-components)
    - Consolidate runs of `TText.text_t` within `TElement`s while parsing. (SEE: consolidating-text-templates)
- Updated `processor.py` to match parser changes.
  - Remove instrumentation because it is now part of parsing.
  - Interpolation is now performed on `TNode`s and their structures.
- Updated `nodes.py` as they now are fully realized during serialization
    - Try to escape text content based on container tag if we have an Element otherwise fallback to markupsafe.
- Consolidated the escaping/formatting into a new module: `tdom/escaping.py`.
- Consolidated the placeholder handling into a new module: `tdom/placeholders.py`
- Add helper for `parser_tests` that helps when comparing `TNode`s because primitives don't work for `Template` or for the consolidated format for dynamic attributes.

#### Notes

- (node-inherits-from-tnode): My thinking here was that it might be desired to just bypass TNodes completely for static content and partially fill `TNode`s with `Node`s.  Right now that isn't happening and another direction might be to strictly forbid `TNode`s and `Node`s be found within the same tree.
- (templates-embedded-in-components): The template embedded in a component can be extracted and passed into the component for it to use but since the attributes cause the "start" of the embedded template to drift away from the actual tag interpolation things get tricky.  I did not have much luck using the line/position information within the parser itself.  There might be a much easier way.  For example `t"<{MyComp} {attrs} selected></{MyComp}>"`, `embedded_t = t""`, then somethings like `invoke_comp(my_comp, resolved_attrs, embedded_t)`.  This isn't super useful but maybe later there would be something like `t"<{MyComp} {attrs} selected></{MyComp}>"` and then maybe within the component something like `interpolate_slots(embedded_t, (default_logo_src,))`.  In this case it is more a template of a template.  
- (lru-cache-with-callable-infos): Not sure what to do about caching the `TNode` WITH the callable infos.  I feel like that is probably a feature I just didn't understand.  Maybe once I understand how it works I could fit it back in.  As I understood it you might want to use a different callable everytime you render the template?
- (special-endtag-handling-in-parser): I'm not 100% sure I understand the start/stop tag special handling in the parser.  I think maybe `handle_startendtag` was added later but what's nice about 3.14 is ... (almost) everything is available now!  It seemed simpler to remove that special handling and split the methods up especially for handling the `<MyComp />` case.
- (consolidating-text-templates): This was easier to deal with when considering how to escape things and where text "was" when building a `Template` but I wasn't sure if really needed the text split up "As it was fed" or what that purpose would be.  The whitespace and other information should be retained.  The calls to `handle_data` are made as soon as there is any data so we would create many Text objects rather than a single TText object with a consolidated `text_t: Template` within it.

### DIFF
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -3,8 +3,6 @@ from string.templatelib import Template, Interpolation
 from collections.abc import Iterable
 from html.parser import HTMLParser
 
-from markupsafe import Markup
-
 from .nodes import (
     VOID_ELEMENTS,
     TComment,


ianjosephwilson:
> ```
> t'<{Comp} />'
> ```
> 
> that's the start-end tag in a nutshell, which looks more aligned with React and less boring than:
> 
> ```
> t'<{Comp}></{Comp}>'
> ```
> 
> which is both error prone and provides nothing to users ... so, if there was any confusion around that, I hope I've explained that in here 👋
> 
> **edit**
> 
> forgot the implicit `</>` closing tag, still aligned with React, less boring and error prone than `t'</{Comp}>'` later on ... or any other tag, to be honest, as that `</>` is an explicit closing tag.

I just *tried* to keep the same behavior but I changed the implementation.  I was just concerned I didn't fully understand the reason for the earlier implementation and maybe the behavior changed in a way I didn't foresee.  The interpolated closing tag version is actually kind of a poor fit because the second interpolation is sort of a "lame duck" value.  "Do we check it?"  "Do we just check the `expression`?"  "What if the endtag value is different from the begintag value?"  I think the anonymous closing tag makes more sense but I didn't try to add that.

davepeck:
> Sorry I'm over here making so much noise.

Not at all! 

Now that I've had a chance to mull it over, I've found this PR to be immensely helpful: in part for the ideas it contains, and in part for how it relates to a PR I started working on before the holidays. So: thank you, thank you! :-)

Stepping back, I think my primary concern about the current `main` branch is that (a) the "semantics" of interpolations aren't explicitly captured; your `TNode` hierarchy starts to tease this apart, and (b) the rules for rendering those interpolations into final values are scattered into a few places. (Related: I think the `main` branch processor currently does too much: placeholder injection/detection, implementing semantic rules, and final rendering.)

**What I like about this direction:**

The introduction of a `TNode` hierarchy that explicitly describes parse structure *before* interpolations are resolved is where I want to go, too. The current `main` branch conflates "what does this template mean?" with "resolve and render it," and I don't like that. Your PR starts to pull those apart.

**Where I'd like to head:**

After thinking through the implications, I *think* I like a two-layer architecture:

1. **TemplateParser** — Takes a `Template` (and no other type aka not `str | t.Iterable[str]`), handles placeholder injection/extraction, parses HTML, and produces a `TNode` tree with explicit semantic types. This is where the *rules* live: "an interpolation in tag position means component invocation," "a bare interpolation in attribute position means spread," etc. The output is cacheable by `Template.strings`.

2. **Processor** — Takes a `TNode` tree plus interpolation values, resolves everything, invokes components, and produces a final `Node` tree.

I originally wondered about splitting HTMLParser out as a separate template-unaware layer &mdash; that is, having `HTMLParser` (which returns `Node`), `TemplateParser` (built on top of it, returns `TNode`), and `Processor` (back to `Node` again) but I don't think there's a real use case for standalone HTML parsing in tdom's world, and the combined approach avoids an extra traversal.

I think the parser itself in the PR intermingles several concerns within its callbacks: placeholder tracking, node construction, attribute classification, component validation, error reporting. I'm not sure exactly what it would look like, but I imagine there's a path to a nicer internal factoring.

I do think `TNode` and `Node` should remain separate hierarchies with different purposes. `TNode` is abstract, cacheable structure. `Node` is concrete, manipulable HTML.

**A few more detailed thoughts:**

*Explicit attribute types*: The implicit tuple shapes for attributes (`(str,)` vs `(str, int)` vs `(str, str, int, str)`) push the "what kind of thing is this?" question into runtime conditionals. I _think_ I'd prefer explicit types like `StaticAttribute`, `DynamicAttribute`, `SpreadAttributes` so the rules are discoverable in the type definitions themselves?

*`Template` with `str | int`*: I went back and forth on the "index templates" that live under the hood and in the `TNode`s. In the end I decided it's confusing to reuse `string.templatelib.Template` when the interpolation values are indices rather than actual values. How do you feel about making this a distinct type (or just store indices directly on the `TNode` fields?)

(As an aside that's maybe moot: in 3.14, `Interpolation` can be constructed with _just_ a value aka `Interpolation(42)`. There are a bunch of places where you provide defaults for expression/conv/format_spec. Also, I think there's some vestigial stuff around `format_spec` in the PR right now?)

*`strings_slice_begin` / `strings_slice_end`*: I see these being set and tested, but not consumed anywhere. If I understand correctly, this is scaffolding for passing a sub-`Template` to components? I didn't quite understand where we're going here...

*Text consolidation*: The old "chunk-boundary-preserving" behavior of the HTMLParser in the `main` branch is vestigial. It's totally fine to combine neighboring bits of text.

---

Not quite sure what the next steps are, but this is a great conversation starter.

One question: are there bits of your PR that I missed or that I'm undervaluing? 

Thanks again for this!

davepeck:
> maybe it was solid?

Full disclosure: I had to go back and remind myself how JSX worked!

I feel like I've run across `</>` (or maybe `<//>`) as more than just "close a fragment" elsewhere. But when I went looking again (solid? lit-html? etc.) I couldn't find it. At least I'm comforted by hallucinating along *with* you. :-)

davepeck:
Closing per discussion.

Issue #77, opened 2025-12-03
Proposal to remove callable infos from cache key


ianjosephwilson:
If we take the callables out of the cache key then the parse result can be reused more frequently and it would make it easier to pre-cache the templates because dummy data could be used.

This was indirectly mentioned in #76 and it is kind of a subtle but still a change.  Do we *need* the callable info to be part of the cache key?  We could AFAIK just use whatever the template gave on the first run and then its just a "help/convenience" to the user to let them know their closing component tag doesn't make sense. 

They were extracted here:
https://github.com/t-strings/tdom/blob/312ce2d9d98165550dd7f55d94007e3b0d0c2b46/tdom/processor.py#L203

Then they are in the cache here:
https://github.com/t-strings/tdom/blob/312ce2d9d98165550dd7f55d94007e3b0d0c2b46/tdom/processor.py#L173

I was thinking during parsing we could just perform a check such as:
```python
if template.interpolations[start_tag_interpolation_index].value != template.interpolations[end_tag_interpolation_index].value:
    raise ValueError('Your start tag and end tag component callables do not match.')
```

I think it might be possible to factor this out and change it before dealing with the PR if you think that would be helpful.

PR #80, opened 2025-12-04
Remove a dependency on the template interpolations from the lru_cache.

ianjosephwilson:
This is a smaller change for #77 to try to simplify #76 .

- Move the callable info computation inside the instrument and parse function so we can still use them for match making the start and end tags of components but they don't interfere with the hashing calculation.
- Add `CachedTemplate` wrapper class to define `__hash__` and `__eq__` for a `Template` to actually change the hashing function and still let the interpolations into the underlying instrument and parse function.
- Removed wrapper function from `_instrument_and_parse_internal` and rename it to `_instrument_and_parse`.  We just wrap the template in `CachedTemplate` now and we can do that in the `html()` function.
- Added a few tests for `CachedTemplate` but as far as I can tell the lru_cache at the center of this isn't very accessible in the tests.  That is something we should probably resolve both for the tests but also if the settings need to be adjusted by a user of the library.

**Note:** That the actual inspection of the call signature of components for interfacing them with the resolved attributes is handled somewhere else and *should not* be affected by this:  [`callable_info = get_callable_info(value)`](
https://github.com/t-strings/tdom/blob/312ce2d9d98165550dd7f55d94007e3b0d0c2b46/tdom/processor.py#L477)

### DIFF
--- a/tdom/processor.py
+++ b/tdom/processor.py
@@ -171,13 +171,22 @@ def _instrument(
 
 
 @lru_cache(maxsize=0 if "pytest" in sys.modules else 512)
-def _instrument_and_parse_internal(cached_template: CachedTemplate) -> Node:
+def _instrument_and_parse(cached_template: CachedTemplate) -> Node:
     """
     Instrument the strings and parse the resulting HTML.
 
     The result is cached to avoid re-parsing the same template multiple times.
     """
     template = cached_template.template
+    # This exists to handle the syntax we've settled on for
+    # component invocation, namely that callables are directly included as
+    # interpolations both in the open *and* the close tags. We need to make
+    # sure that matching tags... match!
+    #
+    # If we used `tdom`'s approach of component closing tags of <//> then we
+    # wouldn't have to do this. But I worry that tdom's syntax is harder to read
+    # (it's easy to miss the closing tag) and may prove unfamiliar for
+    # users coming from other templating systems.
     callable_infos = tuple(
         _callable_info(interpolation.value) for interpolation in template.interpolations
     )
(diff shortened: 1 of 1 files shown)


ianjosephwilson:
Yes, this is something we'd have to check at run-time.  Maybe it would be an appropriate use of `assert` that could be removed in production.  It feels wrong but maybe we should really just ignore it?  I guess the real advantage of checking it is to detect a mistake someone made where that interpolation was *meant* to contain something else and its hard to debug because its silently being ignored.

We still would have to haul the book-keeping around to track it though.  I was using `component_info` to keep this information.  We'd need to pass the information into the parser, then stick it on the `Element` and then check it during interpolation/processing.  This is one of the situations where the "tail wagging the dog" situation with the instrumentation happening outside the parser, before parsing, and then being resolved after parsing outside the parser complicates things.

#### Potential solution

We could do some bookkeeping here and pass it into the parser:
https://github.com/t-strings/tdom/blob/737a7f2d4e26746e778d8bce0113bc93effd7725/tdom/processor.py#L166

Then when we create the element we need to stick that book-keeping onto the Element, in the draft branch I made I put it on `component_info`:
https://github.com/t-strings/tdom/blob/737a7f2d4e26746e778d8bce0113bc93effd7725/tdom/parser.py#L52

Then when we invoke it we could check that the callables are equal:
https://github.com/t-strings/tdom/blob/737a7f2d4e26746e778d8bce0113bc93effd7725/tdom/processor.py#L516

davepeck:
Ah, I see! Thank you. 

Sorry, I've been pretty dumb about this -- your suggestion that we check component matching when *rendering* the result, rather than when injecting placeholders, is great, and allows us to avoid adding component identity to the cache key. (Your nested component function is a pretty compelling example of where caching with identities will fail entirely, anyway.)

Great, agreed, this is the way to go. I gather PR #81 gets us there from the current `main` branch? (But maybe this is better tied to the bigger swing in that other PR?)

ianjosephwilson:
> Great, agreed, this is the way to go. I gather PR #81 gets us there from the current `main` branch? (But maybe this is better tied to the bigger swing in that other PR?)

Exactly, let's wait and see if the bigger swing cleans up a little bit.  If a solution to this is in there then we now have the background for it to not be a surprise.

PR #82, opened 2025-12-08
Parse t-strings into TNodes/TAttributes representation using TemplateParser

ianjosephwilson:
#### Changes

- Added "template" nodes:  `TElement`, `TFragment`, `TComponent`, `TComment`, `TText`, `TDocumentType`
    - Store template text in a `Template` object on `TComment` and `TText` (The "template" equivalents of `Comment` and `Text`).
    - Store template attributes in `StaticAttribute`, `SpreadAttribute`, `InterpolatedAttribute`, `TemplatedAttribute`
    - Added `TComponent` concept.
    - Use `TNode` type alias and `TNodeBase` subclass for "template" nodes.
    - Use `TAttribute` type alias for "template" attributes.
- Evolved `parser.py:NodeParser` to `parser.py:TemplateParser`
    - Externally the parser now considers an entire `Template` rather than being gradually fed chunks.
    - Placeholders are inserted as before but are removed when found during parsing:
        - (1) component start/end tags, (2) attributes, (3) within an element's text, (4) within a comment's text
        - If a placeholder is not found or found in a weird location an error is raised.
    - Grouped `Template` specific parsing state into separate `TemplateState` object.
    - Added `OpenTag` concept to deal with "half-built" open tags until they are closed.
    - Each tag handler dispatches to one of the 3 `OpenTag` concepts.
    - Handle fragments more explicitly.
- Updated `processor.py` to match parser changes.
  - Remove instrumentation because it is now part of parsing.
  - Interpolation is now performed on `TNode`s and their structures.
  - Added component starttag and endtag value check.
- Updated `nodes.py` as they now are fully realized during serialization
    - Try to escape text content based on container tag if we have an Element otherwise fallback to markupsafe.
- Consolidated the escaping/formatting into a new module: `tdom/escaping.py`.
    - Still predominantly use markupsafe's `escape` for escaping.
- Consolidated the placeholder handling into a new module: `tdom/placeholders.py`
- Add helper for `parser_tests` that helps when comparing `TNode`s because primitives don't work for `Template` or for the consolidated format for `TAttribute`s.

### DIFF
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -1,7 +1,9 @@
+import sys
 import typing as t
 from string.templatelib import Template, Interpolation
 from html.parser import HTMLParser
 from dataclasses import dataclass, field
+from functools import lru_cache
 
 from .nodes import (
     VOID_ELEMENTS,
@@ -428,11 +430,35 @@ class TemplateParser(HTMLParser):
         )
 
 
-def parse_html(
-    template: Template,
+@lru_cache(maxsize=0 if "pytest" in sys.modules else 512)
+def _parse_html(
+    cached_template: CachedTemplate,
 ) -> (
     TNode  # @TODO: Might be more consistent for this to always be a container.
 ):
+    parser = TemplateParser()
+    parser.feed_template(cached_template.template)
+    parser.close()
+    return parser.get_node()
+
+
+@dataclass
+class CachedTemplate:
+    """Attempt to cache template just by its strings."""
+
+    template: Template
+
+    def __hash__(self):
+        return hash(self.template.strings)
+
+    def __eq__(self, other):
+        return (
+            isinstance(other, CachedTemplate)
+            and self.template.strings == other.template.strings
+        )
+
+
+def parse_html(template: Template) -> TNode:
     """
     Parse a string, or sequence of HTML string chunks, into a Node tree.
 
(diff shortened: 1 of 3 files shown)


davepeck:
Thanks @ianjosephwilson &mdash; this will take a little while to review, but really appreciate you continuing to work on this!

davepeck (commented):
Overall, I think this is great. It's a huge amount of deeply thoughtful architectural work and I really appreciate it, @ianjosephwilson !

I have lots of detailed thoughts but one possibility is that we merge it into `main` and then I can go and kind of push things a little further in the direction I envision.

I mentioned I put together a separate branch that achieved many of the same goals. I revisited that branch yesterday and took some learnings (and outright took _many_ chunks of code!) from your PR to arrive at #83. I posted that PR mostly to help us continue the conversation. 

Some key differences: I break out parsing state differently, into `PlaceholderState`, and I do not reuse `string.templatelib.Template`, instead preferring a new `TemplateRef` type. The exact factoring for handling interpolations in the processor seems a bit cleaner, in part because of its use of these two types.

davepeck:
> It would be nice to keep the `OpenTag` stuff because it allows for the `TNodes` to be immutable and solves a "kludge" where the `TComponent` is in an incomplete/undefined state until closed (there will probably be other issues like this).

Yes, after dwelling on it, I think we should make `TNode`s immutable and keep the `OpenTag` stuff, wherever we land.

davepeck:
> "is this a single thing or a container" all over the place.

I was feeling a bit uncomfortable about just that question in both this PR and in #83 -- but a problem for the near-future, I suppose. :-)

davepeck:
Per our conversation today, I'm merging this into `main`. Tomorrow I'll update PR #83 to build on top of this.

PR #83, opened 2025-12-10
Architecture update based on conversations with Ian

davepeck:
This is my own take on discussions with Ian. It is intended to continue the conversation.

I started this work before the holidays and wrapped it yesterday &mdash; well, got it to a point where it was worth looking at.

Yesterday I also brought in a number of pieces of Ian's PR #82 (`escaping.py`, `placeholders.py` to some extent, and names for many of the `TNode` types) to kind of figure out what a joint assemblage might look like...

### DIFF
--- a/tdom/processor.py
+++ b/tdom/processor.py
@@ -289,7 +289,7 @@ def _kebab_to_snake(name: str) -> str:
 
 def _invoke_component(
     attrs: AttributesDict,
-    children: list[Node],
+    children: list[Node],  # TODO: why not TNode, though?
     interpolation: Interpolation,
 ) -> Node:
     """


tdom/parser.py
```diff
@@ -41,60 +187,85 @@ def handle_starttag(
...
     def handle_decl(self, decl: str) -> None:
-        if decl.upper().startswith("DOCTYPE"):
-            doctype_content = decl[7:].strip()
-            doctype = DocumentType(doctype_content)
-            self.append_child(doctype)
-        # For simplicity, we ignore other declarations.
-        pass
+        ref = self.placeholder_state.remove_placeholders(decl)
+        if not ref.is_static:
+            raise ValueError("Interpolations are not allowed in declarations.")
```
ianjosephwilson: I think this is a good idea to explicitly take a stance on this.  Even if the situation seems unlikely it will help catch bugs and if someone does try it then the library does not seem broken rather it just doesn't support it.

tdom/parser.py
```diff
@@ -41,60 +187,85 @@ def handle_starttag(
...
+            )
+        doctype_content = decl[7:].strip()
+        doctype = TDocumentType(doctype_content)
+        self.append_child(doctype)
 
     def in_content_element(self) -> bool:
         """Return True if the current context is within a content element."""
         open_element = self.get_open_element()
-        return open_element is not None and open_element.tag in CONTENT_ELEMENTS
+        return (
```
ianjosephwilson: I don't think you are using this method anymore because you are putting the data into a `ref`.

tdom/parser.py
```diff
@@ -1,38 +1,184 @@
...
 )
+from .placeholders import PlaceholderState, TemplateRef
+
+# ------------------------------------------------------
+# Semantic attribute types
+# ------------------------------------------------------
+
+
+@dataclass(slots=True)
+class TLiteralAttribute:
```
ianjosephwilson: I like this name a lot better than StaticAttribute.

tdom/parser.py
```diff
@@ -1,38 +1,184 @@
...
+    name: str
+    value: str | None
+
+
+@dataclass(slots=True)
+class TInterpolatedAttribute:
+    name: str
+    value_i_index: int
```
ianjosephwilson: I experimented with `ip_index` as a shorter name than `interpolation_index`, I just wonder if the standalone version `i_index` is a little confusing.  Anything is better than `interpolation_index` and it would be nice to just use that throughout the code base consistently which looks like you've started here.  I guess another caveat which you don't have in here is sometimes you might need to know the index into `strings` and what would could that be called?

tdom/placeholders.py
```diff
@@ -0,0 +1,121 @@
...
+    """Reference to a template with indexes for its original interpolations."""
+
+    strings: tuple[str, ...]
+    """Static string parts of the original string.templatelib.Template"""
+
+    i_indexes: tuple[int, ...]
+    """Indexes of the interpolations in the original string.templatelib.Template"""
+
+    @property
+    def is_static(self) -> bool:
```
ianjosephwilson: I wonder if this should match the nomeclature of `TLiteralAttribute` with `is_literal()`/`literal()`/etc.

tdom/parser.py
```diff
@@ -1,38 +1,184 @@
...
+        value_ref = (
+            self.placeholder_state.remove_placeholders(value)
+            if value is not None
+            else None
+        )
+
+        # CONSIDER: allow templating a name? A name *and* a value? I mean,
+        # why not?
+
+        if name_ref.is_static:
```
ianjosephwilson: These named properties are *wayyyy* better than the `match` usage.  I felt like I  was trying to wash the beach whenever I tried to cleanup the attribute building in the other branch.

tdom/utils.py
```diff
@@ -86,3 +87,39 @@ def format_interpolation(
         interpolation.conversion,
         formatters=formatters,
     )
+
+
+def template_from_parts(
```
ianjosephwilson: This goes in that new module too.

tdom/parser.py
```diff
@@ -41,60 +187,85 @@ def handle_starttag(
...
         if not self.stack:
             raise ValueError(f"Unexpected closing tag </{tag}> with no open element.")
 
         element = self.stack.pop()
-        if element.tag != tag:
-            raise ValueError(f"Mismatched closing tag </{tag}> for <{element.tag}>.")
+        if isinstance(element, TElement):
+            if element.tag != tag:
+                raise ValueError(
+                    f"Mismatched closing tag </{tag}> for <{element.tag}>."
```
ianjosephwilson: Those snakes pop out here if you don't close a component correctly.  As endearing as they are I feel like that should be an implementation detail and if they get to the user it is a bug.  I think that is something to discuss.  Like I said in one of my other gazillion comments (sorry!) I think that is the big difference between the branches I would like resolved somehow.

tdom/parser.py
```diff
@@ -1,38 +1,184 @@
...
+        tag_ref = self.placeholder_state.remove_placeholders(tag)
+        if tag_ref.is_static:
+            return TElement(tag=tag, attrs=tattrs, children=[])
+        if not tag_ref.is_singleton:
+            raise ValueError(
+                "Component element tags must have exactly one interpolation."
+            )
+        return TComponent(
+            start_i_index=tag_ref.i_indexes[0],
+            attrs=tattrs,
```
ianjosephwilson: This is where I was hoping to store the component's body start "s_index", but I realized I think can probably infer it from the attributes like this:

```python
def find_body_slice(comp):
    for tattr in reversed(comp.attrs):
        i_index = get_last_i_index(tattr)
        if i_index is not None:
            return i_index + 1, component.end_i_index
    return component.start_i_index + 1, component.end_i_index
```

ianjosephwilson (commented):
I added a few notes in here.  I'm fine with the placeholder state being held outside but I think it is just postponing the inevitable.  The TemplateRef seems way better than the match stuff I have.  It would be nice to keep the OpenTag stuff in the other branch so we can make the TNodes immutable (as mentioned there).

tdom/parser.py
davepeck: Yeah, I prefer your `OpenTag` approach.

PR #85, opened 2025-12-12
Performance benchmarks

pauleveritt:
Scripts to generate sample data and run performance benchmarking and profiling, then update docs.

### DIFF
--- a/docs/benchmark.md
+++ b/docs/benchmark.md
@@ -59,13 +59,16 @@ Templates are parsed into an intermediate TNode representation first, then
 processed into the final Node tree:
 
 ```python
+from tdom.parser import TemplateParser
+from tdom.processor import _resolve_t_node  # subject to change!
+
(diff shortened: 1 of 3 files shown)


ianjosephwilson:
@pauleveritt I meant in any of your other projects.  Are you using the component's embedded children or is a component more of just an insert point?  I know it is a feature in other template mechanisms but it does not get used very often.  I'm kind of wondering if we need it.  I think maybe I'll start a discussion around it.

davepeck:
I expect passing children through component invocations (`t"<{Layout}>Hello world!</{Layout}>"`) will be common amongst users of `tdom`. It appears everywhere in _my_ uses, in any case. :-)

(More advanced systems have a notion of slots, but I've avoided that for now.)

pauleveritt:
@davepeck I'll certainly clean things up if you'd like to keep it. But: decide if you want to keep it. [wink]

PR #95, opened 2025-12-19
Remove anonymous fragment for now.

ianjosephwilson:
- Remove anonymous fragments from the parser for now: ie. `<></>`.


pauleveritt:
That's cool that you detect it and raise `NotImplemented`. You removed all mention from the README, I believe. Should we have a quick callout saying "some of you might be looking for....but not needed"?

PR #96, opened 2025-12-19
Wrap up low level placeholder handling into PlaceholderConfig helper

ianjosephwilson:
- Move low level placeholder operations and constants into placeholder config helper class.
- This is to avoid potential free-threading issues but overall we probably shouldn't be using dynamic module level variables anyways.
- Long term we probably want to share the placeholder configuration across a "group" of parsers but I think this is ok for now.

### DIFF
--- a/tdom/placeholders.py
+++ b/tdom/placeholders.py
@@ -1,3 +1,4 @@
+from dataclasses import dataclass, field
 import random
 import re
 import string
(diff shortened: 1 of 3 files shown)


pauleveritt:
Thanks for doing this. Once this is merged, I'll switch my aria-testing package (which I use heavily in other packages) to free threading and see if I spot any problems. I'll also put the pytest-freethreaded stuff in that package and keep trying to trigger.

As a note, the only *potential* (but unlikely) other issue CC found:

#### 🟡 **POTENTIAL ISSUE: Mutable Dataclass Children** (nodes.py, parser.py)
```python
@dataclass
class OpenTElement:
    tag: str
    attrs: tuple[TAttribute, ...]
    children: list[TNode] = field(default_factory=list)  # Mutable!
```

**Problem:** `list` is mutable. If multiple threads access/modify the same OpenTElement's children, race conditions occur.

**Current Safety:** These are created per-parse operation and not shared between threads, so likely safe in practice.

**Impact:** LOW - Only matters if users share parser objects across threads (discouraged pattern)

Issue #89, opened 2025-12-17
Resolve attribute resolution!


davepeck:
We're much closer now. This issue asks us to sit back and consider the final spec we'd like to support for attribute resolution and merging, and then to document and implement it.

pauleveritt:
Cool writeup, thanks. Regarding the point that the browser is actual consumer, I agree that is correct, with the qualifier of "ultimately." 

In my writeup about paths, I had kind of hoped we'd have a place to stash system data on a node. Stuff important during the pipeline, but stripped out during rendering. (Such as path info.) I realize this is premature solution-icing: we might find other ways to do it vs. storing data on nodes.

ianjosephwilson:
Alternatively the last block could be combined like this although as I was saying before usually there a big list of stuff and splitting a conditional like this into two parts would be confusing to come back to:

```python
def get_body_container():
    toggle_classes = {'theme-default': not is_holiday, 'theme-holiday': is_holiday}
    def DefaultContainer(add_classes: list|None=None, toggle_classes:dict|None=None):
        return t"""<div class="theme-default bg-red" class={add_classes} class={toggle_classes}></div>"""
    
    return t"<{DefaultContainer} toggle_classes={toggle_classes}>...</{DefaultContainer}>"
```

And final final,we drop toggling completely and just do lists/strs:

```python
def get_body_container():
    add_classes =['theme-holiday']
    def DefaultContainer(add_classes: list|None=None):
        classes = add_classes[:] if add_classes else []
        if [cn for cn in classes if not cn.startswith('theme-')]:
            classes.append('theme-default')
        # bg-red cannot be removed
        return t"""<div class="bg-red" class={classes}></div>"""
    
    return t"<{DefaultContainer} toggle_classes={toggle_classes}>...</{DefaultContainer}>"
```

Just some more things to consider.

ianjosephwilson:
I made a much more digestible (less experimental) set of changes that I think turned out pretty well in #97 .  There are maybe a few things we would want more strict or less strict but it is pretty close to what is there now ... but with merging.  It washes most of the weird edge cases too which is nice.

ianjosephwilson:
After thinking about this PR a while maybe we should start a bit more strict and then relax things later if there is friction when using it.  So I guess I'm thinking I should adjust it to the `Alt1` here or even going to `Alt1` and `Alt2` and then seeing how things go when using it.

##### Current PR (roughly)

```python
[...]

PR #97, opened 2025-12-26
Refine Attribute Resolution

ianjosephwilson:
Refine attribute resolution to allow for merging of `class` and `style` attrs and smooth out a few edge cases.

### Summary
  - Accumulate `class` values together internally as a `dict[str, bool]` and output to a `str()` or `None` to omit when empty.
  - Accumulate `style` values together internally as a `dict[str, str|None]` and output to a `str()` or `None` to omit when empty.
  - Expand `aria` and `data` as before but do not omit `None` or `False` during intermediary expansion.

#### Major Changes
- `class` changes
    - Merge from left to right except when not forced to merge, ie. pass through singleton literal `class`.
    - If merging is active then always split `str()` values into a list of classes.
    - Don't allow sequences within sequences, ie. `['btn', ['btn-primary']]`
    - Don't allow dicts within sequences, ie. `['btn', {'btn-primary': True}]`
    - Retain `False` values from `dict` form to allow classes previously merged to be "toggled" off.
    - Move class handling into `processor.py`, drop `classnames()` helper.
- `style` changes
    - Merge from left to right except when not forced to merge, ie. pass through singleton literal `style`.
    - If merging is active then always extract `str()` values into property name / value dictionary.
    - Retain properties with `None` value from `dict` form to allow styles previously merged to be "toggled" off.
- `aria` / `data` changes
    - Do not omit attributes with values of `None` or `False` right away so that they may override prior values.
    - Ignore if `None` to allow for `def Comp(img_data=None)` type pass-through optional arguments

#### Types

Just for a quick understanding of how this could look these would be for the non-literal values:
```python
type ClassValueItem = None | str
type ClassValue = ClassValueItem | dict[str, bool] | Sequence[ClassValueItem]
type StyleValue = None | str | dict[str, str | None]
type AriaValue = None | dict[str, object]
type DataValue = None | dict[str, object]

# For example
class_list: ClassValue = ["btn"]
class_dict: ClassValue = {"btn-primary": True}
button_t = t'<button class="theme-holiday" class={class_list} class={class_dict}>Save</button>'
assert str(html(button_t)) == '<button class="theme-holiday btn btn-primary">Save</button>'
```

### DIFF
--- a/tdom/processor_test.py
+++ b/tdom/processor_test.py
@@ -507,9 +507,10 @@ def test_multiple_attribute_spread_dicts():
 
 
 def test_interpolated_class_attribute():
-    class_list = ["btn", "btn-primary", None]
+    class_list = ["btn", "btn-primary", "one two", None]
     class_dict = {"active": True, "btn-secondary": False}
     class_str = "blue"
+    class_space_sep_str = "green yellow"
     class_none = None
     class_empty_list = []
     class_empty_dict = {}
@@ -518,17 +519,18 @@ def test_interpolated_class_attribute():
         t' class="red" class={class_list} class={class_dict}'
         t" class={class_empty_list} class={class_empty_dict}"  # ignored
         t" class={class_none}"  # ignored
-        t" class={class_str}"
+        t" class={class_str} class={class_space_sep_str}"
         t" >Click me</button>"
     )
     node = html(button_t)
     assert node == Element(
         "button",
-        attrs={"class": "red btn btn-primary active blue"},
+        attrs={"class": "red btn btn-primary one two active blue green yellow"},
         children=[Text("Click me")],
     )
     assert (
-        str(node) == '<button class="red btn btn-primary active blue">Click me</button>'
+        str(node)
+        == '<button class="red btn btn-primary one two active blue green yellow">Click me</button>'
     )
 
 


ianjosephwilson:
Not quite confident yet, I think we'd need to merge them and then try them out a bit and see if they are annoying to use.  I'm not sure what format would be best overall since there are multiple things all intermingled together to get to the end result.

pauleveritt:
@ianjosephwilson Looks like a policy I should get into the specification stuff I was doing for @koxudaxi ... if you can jot some thoughts down in a reply about styles or other attribute edge cases, I'll capture them.

ianjosephwilson:
@pauleveritt Do you mean the rules for how to validate a t-string used by tdom ?

ianjosephwilson:
@davepeck Later this week when you get a chance I think this is a strong contender for merging but look it over.  I'm going to put another note in the related issue.  I just had it drafted because it was stacked on top of the other small PR you merged earlier.

davepeck:
@ianjosephwilson Thanks so much for this PR and for your patience while I, uh, put my feet up on a tropical beach for a couple weeks. :-)

After reading this over in detail, I think it's a great stake in the ground and I'm happy to merge it and push an update to PyPi. My instinct is that we'll need to use it for a while to see how we feel about the behavior. I'm happy to make adjustments going forward, including large breaking changes (still pre-alpha!), as we learn more. 

Does this sound good to you? If you're feeling good about the PR, I'll go ahead and merge.

PS: good riddance to `classnames()`. 😅

ianjosephwilson:
Sorry I keep stacking commits on.  This is a little bit more restrictive, as mentioned in #89 , but let's try this out and if we keep hitting cases where we need the other forms we can add them back in.  A good overview is in `def test_interpolated_class_attribute()`.

PR #109, opened 2026-03-17
Prefactor some parts to reduce size of other PR

ianjosephwilson:
- new modules
  - start preparing for more html spec constants by creating new module: `htmlspec.py`
  - move html dunder protocol into a module so we can more easily share it: `protocols.py`
  - Add sentinel for use in the project (until Python standardizes one, in 3.15? maybe): `sentinel.py`
- `TemplateRef` additions
  - add helper method for TemplateRef that iterates parts, omitting empty strings, like `Template.__iter__` does.
  - add helper method to TemplateRef that resolves the interpolation values from a paired `Template` object in between the `TemplateRef` `strings` based on the `i_indexes`.
- new feature: `"callback"` format_spec triggers new formatter
  - Add callback formatter that executes a zero-arg callable before resolving the template values.  This would be opt-in instead of any Callable passed into the `Template` being called automatically.

### DIFF
--- a/tdom/format.py
+++ b/tdom/format.py
@@ -29,7 +29,11 @@ def convert[T](value: T, conversion: t.Literal["a", "r", "s"] | None) -> T | str
 type FormatMatcher = t.Callable[[str], bool]
 """A predicate function that returns True if a given format specifier matches its criteria."""
 
-type CustomFormatter = t.Callable[[object, str], str]
+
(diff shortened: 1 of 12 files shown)


davepeck (commented):
This seems like a good pre-factor. But I'm a little unclear on where we're heading with component invocation?

PR #112, opened 2026-04-03
Revamp Processor

ianjosephwilson:
* Revamp Processor To Be Template Based
  - Enumerate interpolation cases to simplify logic and establish framework to hone correctness, specially around text types.
* Remove Nodes
* Move `svg` support into processor.
* ~~Add in system context, parent tag context and namespace context.~~ Needs another PR.
* Change Component Signature(s)
```python
def FunctionComponent(children: Template, **kwargs) -> Template:
    return t""
def FactoryComponent(children: Template, **kwargs) -> Callable[[], Template]:
    def ComponentCallable() -> Template:
        return t""
    return ComponentCallable
```
* Change Component Children Processing Order
  - Children is a `Template` so the actual contents are not processed until *after* the component is invoked.
* Change Content Interpolation Rules
  - Zero-arg functions no longer have special treatment:
    - `to_html(t'<div>{(lambda: "dynamic")}</div>') != '<div>dynamic</div>'`
    - A new formatter named `callback` can be used which then interpolates the return value
      - `to_html(t'<div>{(lambda: "dynamic"):callback}</div>') == '<div>dynamic</div>'`
  - ~~Boolean values no longer have special treatment~~ We'll keep this for now.
    - True/False are both ignored.


tdom/escaping.py
```diff
@@ -2,17 +2,25 @@
...
+def escape_html_comment(text: str, allow_markup: bool = False) -> str:
     """Escape text injected into an HTML comment."""
     if not text:
         return text
+    elif allow_markup and isinstance(text, HasHTMLDunder):
+        return text.__html__()
+    elif not allow_markup and type(text) is not str:
+        # text manipulation triggers regular html escapes on Markup
+        text = str(text)
+
```
davepeck, line 26: I might write this preamble as:

```python
if not text:
	return text
if allow_markup and isinstance(text, HasHTMLDuner):
	return text.__html__()
text = str(text)
```

tdom/parser.py
```diff
@@ -364,16 +337,13 @@ def feed_template(self, template: Template) -> None:
...
         """
         Parse a Template containing valid HTML and substitutions and return
         a TNode tree representing its structure. This cachable structure can later
         be resolved against actual interpolation values to produce a Node tree.
-
-        Pass ``svg_context=True`` for SVG fragments that have no ``<svg>``
-        wrapper, so that tag and attribute case-fixing applies from the root.
         """
-        parser = TemplateParser(svg_context=svg_context)
+        parser = TemplateParser()
```
davepeck, line 346: I'm curious about your instinct here. SVG fixes feel very much like a parser-layer concern to me, not a processor -- by the time we reach our processor, we should have dealt with the annoyingness of the base HTML parser we're using.

(**UPDATE** ah, I get it. See further comments.)

tdom/processor.py
```diff
@@ -445,8 +386,8 @@ def _kebab_to_snake(name: str) -> str:
 def _prep_component_kwargs(
     callable_info: CallableInfo,
     attrs: AttributesDict,
-    system_kwargs: dict[str, object],
-):
+    system_kwargs: AttributesDict,
```
davepeck: Perhaps let's drop `system_kwargs` for now and just use `children`?

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+    return "".join(
+        (f' {k}="{escape(v)}"' if v is not None else f" {k}" for k, v in html_attrs)
+    )
 
-    Component invocation rules:
 
-    1. All arguments are passed as keywords only. Components cannot require
-    positional arguments.
+def _fix_svg_attrs(html_attrs: Iterable[HTMLAttribute]) -> Iterable[HTMLAttribute]:
+    """
```
davepeck, line 426: Looking over your SVG-in-processor implementation, it does seem like the better approach. Namely, it allows us to think of SVG serialization as an output concern, which allows us to dynamically decide output format. And `<foreignObject>` is... wow, what a "fun" wrinkle. Okay then!

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+        )
 
-    If no children exist but the callable accepts "children", we pass an
-    empty tuple.
 
-    3. All other attributes are converted from kebab-case to snake_case
-    and passed as keyword arguments if the callable accepts them (or has
-    **kwargs). Attributes that don't match parameters are silently ignored.
-    """
+type FunctionComponentProto = Callable[..., Template]
```
davepeck: I'm okay with dropping the `Proto` suffix for all of these; for some reason, feels more in line with other python type names I've seen here.

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
-    # The cast avoids `call-top-callable` from ty; there *must* be a smarter
-    # way, though... -Dave
-    result = t.cast(Callable[..., t.Any], value)(**kwargs)
-    return _node_from_value(result)
 
+type NormalTextInterpolationValue = (
+    None | str | Template | Iterable[NormalTextInterpolationValue] | object
+)
+# Applies to both escapable raw text and raw text.
+type RawTextExactInterpolationValue = None | str | HasHTMLDunder | object
```
davepeck: Plausible alternative? `RawTextSafeInterpolationValue` and `RawTextAnyInterpolationValue`

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+    # None means unknown not just a missing value.
+    ns: str | None = None
+
+    system: dict = field(default_factory=dict)
+
+    def copy(
+        self,
+        ns: NotSet | str | None = NOT_SET,
+        parent_tag: NotSet | str | None = NOT_SET,
+        system: NotSet | dict = NOT_SET,
```
davepeck: Perhaps let's drop `system` for now, too.

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+    ) -> str:
+        return "".join(self.process_template_chunks(root_template, assume_ctx))
+
+    def process_template_chunks(
+        self, root_template: Template, assume_ctx: ProcessContext | None = None
+    ) -> Iterable[str]:
+        if assume_ctx is None:
+            # @DESIGN: What do we want to do here?  Should we assume we are in
+            # a tag with normal text?
+            assume_ctx = make_ctx(parent_tag=DEFAULT_NORMAL_TEXT_ELEMENT, ns="html")
```
davepeck, line 546: This feels like a fine default to me, at this point?

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+        root = self.parser_api.to_tnode(root_template)
+
+        bf: list[str] = []
+        q: list[WalkerProto] = [
+            self.walk_from_tnode(bf, root_template, assume_ctx, root)
```
davepeck: So... here is where I personally hit a wall.

At a high level: writing an iterative version of a recursive algorithm is fine. For `tdom`, it might one day be necessary if our users ultimately have absurdly deep nestings of templates. (Python's default limit is 1000, I think; hard to imagine many web projects hitting this, even if deeply nested.) This code will almost certainly be a performance win on day one.

But the specific approach in this PR strikes me as quite complex.

First, `walk_from_tnode` has very high cyclomatic complexity. It seems hard to test anything in isolation; AFAIK, we don't: we prefer going through `html()` every time. But if there are bugs where, say, a sub-walker is never re-entered, I don't know if we'd catch them.

The deeper difficulty for me comes from all the shared state. `process_template_chunks` has a `q` and a mutable buffer (`bf`) that's sent downward. `walk_from_tnode` has its own entirely different `q`. And that inner `q` is half-consumed, only to get back to work when the outer "driver" resumes the generator.

There's also some oddity here with `EndTag` in the inner queue; do we use the `ProcessContext` for `EndTag` entries? Do we plan to?

I'm not quite sure what the right answer is here. It feels like there's probably a way to separate "what to do with a node" from "how to schedule the work". The naive recursive variant doesn't suffer these problems because the call stack handles "scheduling the work" implicitly.

Still considering whether the juice is worth the squeeze.

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+    ) -> None | WalkerProto:
+        body_start_s_index = (
+            start_i_index
+            + 1
+            + len([1 for attr in attrs if not isinstance(attr, TLiteralAttribute)])
+        )
+        start_i = template.interpolations[start_i_index]
+        component_callable = t.cast(ComponentCallableProto, start_i.value)
+        if start_i_index != end_i_index and end_i_index is not None:
+            # @TODO: We should do this during parsing.
```
davepeck, line 710: Yeah, this feels post-hoc at the moment. I guess `TComponent` would want to store a children template directly, which gets built by the parser when closing the component? And I suppose `extract_embedded_template` could move to some utility location?

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+        if content is None or content == "":
+            return
+        else:
+            bf.append(
+                self.escape_html_text(
+                    content,
+                )
+            )
+
+    def _process_normal_text(
```
davepeck, line 809: Can we collapse `_process_normal_text()` and `_process_normal_text_from_value()`?

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+                # DO NOTHING
+                return
+            result_root = self.parser_api.to_tnode(result_t)
+            return self.walk_from_tnode(bf, result_t, last_ctx, result_root)
+        else:
+            raise TypeError(f"Unknown component return value: {type(result_t)}")
+
+    def _process_raw_texts(
+        self,
+        bf: list[str],
```
davepeck: The triple of `bf`, `template`, and `last_ctx` feel like their own type &mdash; `RenderState` maybe? Although I don't know if it's actually worth making a refactor here.

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+        elif value is None:
+            # @DESIGN: Ignore None.
+            return
+        else:
+            # @DESIGN: Everything that isn't an object we recognize is
+            # coerced to a str() and emitted.
+            bf.append(self.escape_html_text(value))
+
+
+def resolve_text_without_recursion(
```
davepeck, line 862: This is another method whose intent I think I understand but whose practice is less clear to me at the moment. Specific details to follow...

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+        value = format_interpolation(template.interpolations[content_ref.i_indexes[0]])
+        value = t.cast(RawTextExactInterpolationValue, value)  # ty: ignore[redundant-cast]
+        if value is None:
+            return None
+        elif isinstance(value, str):
+            return value
+        elif isinstance(value, HasHTMLDunder):
+            # @DESIGN: We could also force callers to use `:safe` to trigger
+            # the interpolation in this special case.
+            return Markup(value.__html__())
```
davepeck, line 883: Is there a reason we return trusted `Markup` in the singleton case, but raise a `ValueError` in the other case?

tdom/processor.py
davepeck: As a high-level comment, it took me a moment to understand that the "no recursion" guarantee isn't enforced by the function; it's up to callers to invoke it at the right time.

tdom/processor.py
```diff
@@ -474,151 +415,560 @@ def _prep_component_kwargs(
...
+            value = format_interpolation(template.interpolations[part])
+            value = t.cast(RawTextInexactInterpolationValue, value)  # ty: ignore[redundant-cast]
+            if value is None:
+                continue
+            elif (
+                type(value) is str
+            ):  # type() check to avoid subclasses, probably something smarter here
+                if value:
+                    text.append(value)
+            elif not isinstance(value, str) and isinstance(value, (Template, Iterable)):
```
davepeck, line 906: I assume `not isinstance(value, str)` is because `str` is `Iterable`. The singleton path handles this via an explicit `isinstance(value, str)` earlier in its if-thens.

tdom/processor.py
davepeck: Revisiting this: Your `@DESIGN` comment feels important to me here, since this _appears_ to be the only place in `resolve_text_without_recursion()` where we silently wrap content; where we take on any responsibility for thinking about escaping/safety.

davepeck (commented):
Thanks @ianjosephwilson for this updated PR! I hope I've done a good first-level job at explaining my perspective here...

tdom/processor.py
davepeck: (But feels like it can wait for a future point wise PR.)

davepeck:
One last comment:

> Boolean values no longer have special treatment

I think we need to support this, mostly for familiarity from anyone arriving from Javascript-land. `{someCondition && <div>...</div>}` is an extremely common idiom.

tdom/escaping.py
ianjosephwilson: This seems logically equivalent?  I think maybe I have a tendency to come back later and misread the conditions broken like this and it forces reading the condition and body BUT I looked into this function and it is a case study of different smells so I think I need to revisit it anyways then we can reconsider.

tdom/parser.py
ianjosephwilson: I tried to explain it before but I think it was not concise enough: https://github.com/t-strings/tdom/pull/102#issuecomment-3922522928

>  keep the tnodes as normalized as possible, ie. all lower case as parsed. The semantic information that the SVG is in there OR NOT is already captured. Fixing the case OR NOT seems like another stage in the pipeline


> put the "fixing" into the processor until we have more information about the MathML and larger tdom changes have settled

tdom/processor.py
ianjosephwilson: Is there an alternative proposal for this?  It seems necessary at some point.

tdom/processor.py
ianjosephwilson: This is a weird case but I think "Safe" is not quite right here.  These names were meant more to be exhaustive and less to be friendly.  The suffix is also distracting.  Maybe "Mixed" would be a better parallel to "Exact", or "Mixed" and "Singleton" from `TemplateRef`? `ExactRawText...` vs `MixedRawText...`?  Let me think about it.  `RawTextInexactInterpolationValue` ... obviously ... terrible... lol, "NotTheOtherOneInterpolationValue" would have probably been better.

tdom/processor.py
ianjosephwilson: I guess I just don't see how we can use `tdom` without this and not having to prop drill a `request` into every component at least for webdev.

tdom/processor.py
ianjosephwilson: I think the `ctx` stuff with the `parent_tag` and `ns` is still a little uncooked.   It survived the `svg` stuff but it might need adjusting. But I should try to put this into a more formal spot that puts a stake in the ground that says this-is-our-declared-default rather than this haphazard conditional. 

To clarify though if you did something like `html(t"<clipPath />")` we would spit out `"<clippath></clippath>"` because we can't know you meant to process a SVG fragment as the topmost template.  Whereas something like `html(t"<svg>{t'<clipPath />'}</svg>")` should just work.  Does that make sense?  This is sort of why a `svg()` might still be necessary, OR overriding this default `ctx`, but only in very rare cases.  I'm not sure those cases even really would make sense TBH.

tdom/processor.py
davepeck: (see below)

tdom/processor.py
davepeck: hahaha, yeah. Honestly, I think your names are about as good as we're likely to get.

tdom/processor.py
davepeck: > This is sort of why a svg() might still be necessary

I'm fine with requiring use of a separate `svg()` in the case of `t"<clipPath />"`. Seems like the simplest approach?

tdom/processor.py
davepeck: I've been looking at [`htpy`'s approach to context](https://htpy.dev/usage/#passing-data-with-context) and I think I like it. How and whether it cleanly slots in to the tdom cinematic universe, I'm less sure.

tdom/escaping.py
davepeck: It is logically equivalent, but I personally find it easier to grok. Definitely a minor comment.

tdom/processor.py
ianjosephwilson: I'll have to look at what I did in that parser proposal but I think we can also keep the string indexes that reference the parent template.  Maybe it would be a specialized structure like `TemplateSlice` or something that holds the pointers to the slice in the parent template.  This is working it just might get broken because it is sort of "unofficial".

tdom/processor.py
ianjosephwilson: This is helpful feedback Dave.  I've written and rewritten the processor so many times now I forgot how complicated this would look.  I do feel like there are multiple recursive loops and that is the cause of some of the complexity.  For now though I do want to resolve the other issues and maybe I could take another stab at it later.

I started a rework BACK to recursive and it seems like it will workout OK.  I'll probably finish it this week and we can look at it again.  Thanks for taking the time to look through this.

tdom/processor.py
ianjosephwilson: I think we could maybe delegate from one to the other but I think it is important to have the "entrypoint" function that  (should have) has a cast in it.  I will try to improve this.

tdom/processor.py
ianjosephwilson: Yeah I think maybe its too early here, especially since `last_ctx` itself might get split up.

tdom/processor.py
ianjosephwilson: I think maybe that but in combination with also trying to disambiguate between "true" strings made with `str()` versus subclasses like something from `markupsafe`.  Although this might be an "optimization gone unreadable", I have to go back and review it, and if that is the case I'd probably take it out unless it is related to constraining the type signatures of the interpolations.

tdom/processor.py
ianjosephwilson: This function resolves the interpolations within a template into a string but it does not allow things that "recurse" during processing: templates and iterables.  We don't allow those for scripts, styles, titles, textareas, etc. because it is kind of a edge case and does not really make sense for them to have children:

- normal text (all the other tags that can contain text: div, p, etc.) - entities work and children usually make sense
- raw text (script, style, at least) - entities don't work here and children don't make sense
- escapable raw text (textarea, title, at least) - entities work here but children don't make sense

tdom/processor.py
ianjosephwilson: We can't escape the closing tags in chunks, `chunks=["var x <", "/script>"]` from `t"<script>var x <{'/script>'}</script>"` like normal text.   So we only allow "plain" values EXCEPT when the interpolation is the entire content, ie. an exact match (no chunks): `t"<script>{user_escaped_this_dunder_html_object}</script>"`.

ianjosephwilson:
> One last comment:
> 
> > Boolean values no longer have special treatment
> 
> I think we need to support this, mostly for familiarity from anyone arriving from Javascript-land. `{someCondition && <div>...</div>}` is an extremely common idiom.

Fine, I think it is a small mistake, but we can hopefully make it configurable later on.

ianjosephwilson:
Some diagrams showing the call structures:

### TElement, TFragment and TComponent
```mermaid
graph TD;
    process_tnode-->_process_element;
    _process_element-->process_tnode;
    process_tnode-->_process_fragment;
    _process_fragment-->process_tnode;
    process_tnode-->_process_component;
    _process_component-->process_tnode;
```
### TText (Raw) and TComment
```mermaid
graph TD;
    process_tnode-->_process_raw_texts;
    _process_raw_texts-->resolve_text_without_recursion;
    process_tnode-->_process_escapable_raw_texts;
    _process_escapable_raw_texts-->resolve_text_without_recursion;
    process_tnode-->_process_comment;
    _process_comment-->resolve_text_without_recursion;
```
### TText (Normal)
```mermaid
graph TD;        
    process_tnode-->_process_normal_texts;
    _process_normal_texts-->_process_normal_text;
    _process_normal_text-->_process_normal_text_from_value;
    _process_normal_text_from_value-->_process_normal_text_from_value;
    _process_normal_text_from_value-->process_tnode;
```

### Node Processor
```mermaid
graph TD;        
    _resolve_t_node-->_substitute_and_flatten_children;
    _substitute_and_flatten_children-->_resolve_t_node;
    _resolve_t_node-->_invoke_component;
    _resolve_t_node-->_resolve_t_text_ref;
    _resolve_t_text_ref-->_node_from_value;
    _invoke_component-->_node_from_value;
    _node_from_value-->_node_from_value;
    _node_from_value-->_resolve_t_node;
```

pauleveritt:
I've moved my stuff away from Node and over to this PR. Including context. I currently have around an 80 line "fork" override to enable calling components. 

If I could get a "hook" plug point, I could eliminate it. I (with Claude) have a writeup I can provide.

ianjosephwilson:
@pauleveritt What do you mean "enable calling components"?  I think components should still work, although this roller coaster of pull-requests has been changing a lot.  I'm sorry.  I have been thinking about your attribute rewriting situation.  It is similar to the "svg" attr fix situation in a way and there seems to be emerging behavior that resembles these sort of single level (1-node/1-node's children) change of behavior or multiple level (subtree) change in behavior.  I think there should be some sort of pluggability in there but I'm not sure what/where at this point.  Maybe that is for another PR ?

tdom/processor.py
ianjosephwilson: I forgot about the contextvars I was using before.  I can use that for now I guess at least for a super-global garbage-barge type setup: `request` object with everything accessible off it.  We would need to make it so components can "set" contextvars for their subtree though for it to be generally useful, or at least to reduce the exposure of "magic" globals.  Although I'm not sure there is a way to make it more "software engineering" friendly than that.  I wonder if we could go one step further and instead of just hardcoding contextvar handling into the "middleware" we could just allow a general context manager to be setup by a component before its children are rendered.  I'll have to think about that.  The dataclass/factory style components would provide the context values to set for this to work in between building the component object and calling the component object to get the template.

This a POC I tried to explain a while ago that does the hard-coded contextvars type implementation:
- PROCESSOR 
  - [component signals it wants to set context vars](https://github.com/ianjosephwilson/tdom/blob/ian/revamp_processor_context_poc/tdom/processor.py#L727)
  - [context vars are set while component's children are processed](https://github.com/ianjosephwilson/tdom/blob/ian/revamp_processor_context_poc/tdom/processor.py#L739)
- USAGE
  - [component subtree is wrapped in context setting component](https://github.com/ianjosephwilson/tdom/blob/ian/revamp_processor_context_poc/tdom/processor_contextvars_test.py#L46) 
  - [components in subtree can access contextvar](https://github.com/ianjosephwilson/tdom/blob/ian/revamp_processor_context_poc/tdom/processor_contextvars_test.py#L39)

pauleveritt:
I have a DI system that changes the way components are called. I'd like to configure a function I can provide that handles the calling of a component. The function should be provided the target and have access to the contextvar.

I have a long analysis of this (via an agent) with a lot more detail.

No rush, as my "fork" is now rebased on this PR.

tdom/processor.py
pauleveritt: As FYI, my thing is based on svcs so I have a global registry and per-render containers. svcs doesn't (yet) have nested containers so nesting a container at component boundary isn't something I'm doing. (My previous system had this.)

ianjosephwilson: @davepeck I guess if we don't provide a pattern for certain common integrations then "users" are going to do it anyways and we'll be supporting whatever those ways are.  I think I'd probably just start injecting request and using it because it is easier to migrate and then I might come back and do something more "engineered" later on.  If the option isn't possible it kind of becomes an annoyance.  Like the notorious mouse with only one button.  I think we should put something-like-it in the next PR.  I guess one issue might be how would we tell the tools and I was thinking it might be possible with `Annotated`.
@pauleveritt ok, that makes sense, its kind of a weird case that might be more appropriate for frontend, ie. a multi-level registry versus global

tdom/processor.py
ianjosephwilson: I think we might consider dropping it and just providing documentation for it because it would be so obscure and people will think they need it instead of just combining the fragments into a larger template.  Maybe consider removal in another PR though.

ianjosephwilson:
@davepeck I think this is more digestible compared to the prior solution.  I tried to address some of comments and some of the changes I'll try to put in additional PRs, like making the component processor pluggable.  I think we are getting close to almost having a semi-usable type map.

tdom/processor.py
davepeck: > if we don't provide a pattern for certain common integrations then "users" are going to do it anyways and we'll be supporting whatever those ways are

That's true, and a good note of caution.

I like your POC, and it _does_ seem like tying context var setting to components makes the most sense -- although I imagine there could probably be a generic "ContextVarSetter" component that takes arbitrary attributes and implements `ComponentContextProvider`. 

I dunno, none of it feels quite simple enough from a dev-user perspective yet and I don't have a coherent answer. Are you comfortable kicking this can post this PR, though?

tdom/processor.py
davepeck: > might consider dropping it 

That works for me.

tdom/processor.py
davepeck: Yeah, feels like a post-PR issue to clean up.

davepeck:
@ianjosephwilson Thank you yet again for the big round of updates. I'm feeling much more comfortable with this version of the code. Where are you with this PR? It feels close to mergable to me at this point.

ianjosephwilson:
@davepeck Yeah I think we can merge it and then start making PRs against it.  The `Node` change + component signature change are so big that it kind of is blocking everything else.  It would be nice to get it in so we can start cleaning up things and start putting together prototypes for some of the other features that are needed.

davepeck:
Let's go!

PR #113, opened 2026-04-20
Pluggable Component Processor

ianjosephwilson:
Draft of a pluggable component processor

- factor the "core" invocation of a component out into its own callable object that can be replaced
    - define protocol
    - provide default implementation
- basic test of replacing the default and using a `ContextVar` to provide a global value

### DIFF
--- a/tdom/processor.py
+++ b/tdom/processor.py
@@ -528,14 +528,13 @@ class IComponentProcessor(t.Protocol):
 
     def process(
         self,
-        # TODO: ? processor_api: IProcessor,
         template: Template,
         last_ctx: ProcessContext,
-        component_callable: ComponentCallable,
+        component_callable: t.Annotated[object, ComponentCallable],
         attrs: tuple[TAttribute, ...],
         component_template: Template,
         provided_attrs: tuple[Attribute, ...] = (),
-    ) -> ComponentObject | Template:
+    ) -> tuple[Template, ComponentObject | None]:
         """
         Process available component details into a queryable object or template.
         """
@@ -549,14 +548,13 @@ class ComponentProcessor(IComponentProcessor):
 
     def process(
         self,
-        # @TODO: ? processor_api: IProcessor,
         template: Template,
         last_ctx: ProcessContext,
-        component_callable: ComponentCallable,
+        component_callable: t.Annotated[object, ComponentCallable],
         attrs: tuple[TAttribute, ...],
         component_template: Template,
         provided_attrs: tuple[Attribute, ...] = (),
-    ) -> ComponentObject | Template:
+    ) -> tuple[Template, ComponentObject | None]:
         """
         Process available component details into a queryable object or template.
 
@@ -573,7 +571,28 @@ class ComponentProcessor(IComponentProcessor):
             children=component_template,
             provided_attrs=provided_attrs,
         )
-        return component_callable(**kwargs)
+        res1 = component_callable(**kwargs)  # ty: ignore[call-top-callable]
+        # This integration API seems a lot cleaner but we lose the ability for
+        # component_object.__call__ to be wrapped in any sort of context setting
+        # mechanism provided by the class, but maybe that's ok?  It could be
+        # cached on the instance although that is kind of gross.
+        if isinstance(res1, Template):
+            return res1, None
+        elif callable(res1):
+            res2 = res1() # ty: ignore[call-top-callable]
+            if isinstance(res2, Template):
+                # @TODO: It seems like we should not need this.
+                # Although our check against res2 doesn't seem to affect the
+                # return value of res1.
+                return res2, t.cast(ComponentObject, res1)
+            else:
+                raise TypeError(
+                    f"Component object must return Template when called: {type(res2)}"
+                )
+        else:
+            raise TypeError(
+                f"Component callable must return Template or Callable: {type(res1)}"
+            )
 
 
 class ITemplateProcessor(t.Protocol):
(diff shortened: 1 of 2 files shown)


pauleveritt:
Thanks for looking at this. One other part of what my branch had done: override component location.

Meaning, a template says it wants `Header,` which it imported as `from the_base_app.Header`. But in my "site", I want to replace `Header`. So I have an `IComponentProcessor` that is also able to find an alternative implementation.

To do so, all the custom `IContextProcessor` needs is a context.

Want me to supply some tests to show this?

AhnafCodes:
Also it would be helpful if tdom provides some sort of a Context mechanism (built on contextvars) to share data deeply through the component tree without prop drilling(as React does, like "create_context").
This will help us with common scenarios like:
 - Translations/i18n
 - Error Messages like with Django, 
 - Theming

Something like this could be achieved with it:
```python
from tdom import create_context, html

# Create a context object
ThemeContext = create_context("theme", default="light")

def ThemedButton(text: str):
    # Read the current context value
    theme = ThemeContext.get()
    return html(t"<button class='btn-{theme}'>{text}</button>")

def App():
    # Provide a value for the context
    with ThemeContext.provide("dark"):
        return html(t"<{ThemedButton} text='Click Me' />")

# result = html(t"<{App} />")
# <button class='btn-dark'>Click Me</button>
```

ianjosephwilson:
> Thanks for looking at this. One other part of what my branch had done: override component location.
> 
> Meaning, a template says it wants `Header,` which it imported as `from the_base_app.Header`. But in my "site", I want to replace `Header`. So I have an `IComponentProcessor` that is also able to find an alternative implementation.
> 
@pauleveritt  I think you can just shove everything into a global `ContextVar`.  I tried to flush out a very simplistic dependency injection mechanism that just replaces protocols with a pre-registered implementation.  It is in the `processor_test.py`.  `pyright` isn't happy with some of the typing, ie. using a `Protocol` as a dict key, but I think it seems to work....
    
@AhnafCodes You can use `contextvars.ContextVar` outside of the processor and then pull the values into the component processor but there is no way yet to set a `contextvars.ContextVar` in the "middle" of processing, like to set a context var during a subtree of the component tree.  In this way `tdom` isn't really aware of the contextvar at all: it doesn't set or get them.

pauleveritt:
You have a note `@NOTE: What should this be?` presumably asking what should be the type hint, or shape of the `ComponentCallable` type hint.

This might be an interesting place to justify a refactor. Instead of everything in one big `process` method, perhaps break into a lifecycle with a couple of methods. If nothing else, it would promote re-use by letting people override smaller parts of the default.

In this case, it could let you use a generic. Imagine a `def locate_implementation(self, component_type)` method. The "kind of thing" you asked for could be the type of the return value, using a generic.

I apologize, I lost the plot on `ProcessContext` and that class doesn't have a docstring. With what I was doing, the context was created/destroyed on each top-level render. Roughly meaning, web request. Thus, the context was like a mutable transaction, and subcomponents could add stuff (such as, static assets to go in the `<head>`. )

At the end of the "transaction" (web request), you could gather all that up and generate the `<head>` component.

Is `ProcessContext` like that?

pauleveritt:
Also apologies if I missed the explanation on this: "The `provided_attrs` can be used by extensions that want to provide kwargs even if they are not specified in a template." Can you give a use case?

ianjosephwilson:
> You have a note `@NOTE: What should this be?` presumably asking what should be the type hint, or shape of the `ComponentCallable` type hint.
> 
> This might be an interesting place to justify a refactor. Instead of everything in one big `process` method, perhaps break into a lifecycle with a couple of methods. If nothing else, it would promote re-use by letting people override smaller parts of the default.

I guess I think this is the small part that is a part of the larger lifecycle.  This seemed like the part that would be the most pluggable and still be able to leverage the rest of the library.

> 
> In this case, it could let you use a generic. Imagine a `def locate_implementation(self, component_type)` method. The "kind of thing" you asked for could be the type of the return value, using a generic.
> 
I started looking at generics to solve the custom context situation but you're exactly right.  I have been struggling to understand what the "thing" is that maps the types because `Template` is loose with types but that is conceptually what is going on right now.  I'm not sure how that would affect performance and/or look but I think that is exactly what is going on here (using the `=` to set the "default" type for each `TypeVar`):

```python

## CONCEPT ONLY --  obviously I'd make up a complicated name for each individual TypeVar for realworld usage.
class TemplateProcessor[
    T1=ComponentInterpolationValue,
    T2=NormalTextInterpolationValue,
    T3=RawTextExactInterpolationValue,
    T4=RawTextInexactInterpolationValue,
    T5=InterpolatedAttributeValue, 
    T6=TemplatedAttributeValue,
    T7=ClassAttributeValue,
    T8=StyleAttributeValue,
    T9=AriaAttributeValue,
    T10=DataAttributeValue, 
    T11=SpreadAttribute,
    ](ITemplateProcessor[T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11]):
    # ...


customized_processor = TemplateProcessor[T1=InjectedComponentInterpolationValue](..., component_processor_api=injecting_component_processor, ...)
```

> I apologize, I lost the plot on `ProcessContext` and that class doesn't have a docstring. With what I was doing, the context was created/destroyed on each top-level render. Roughly meaning, web request. Thus, the context was like a mutable transaction, and subcomponents could add stuff (such as, static assets to go in the `<head>`. )

I **was** trying to shoehorn the custom context into that process context and it comes out really confusing, at least for me.  I think it might seem gross to have more "pass-through" arguments all over the place but it is a lot simpler to keep the custom context separate.  I was going to call it `app_ctx` for application context and the `ProcessContext` would be more about where the processor is within the "tree" of `TNodes` and `Templates`.  I don't think it *has* to exist but it seems like there will be demand for it because it is so easy to understand.

ianjosephwilson:
> Also apologies if I missed the explanation on this: "The provided_attrs can be used by extensions that want to provide kwargs even if they are not specified in a template." Can you give a use case?

@pauleveritt I added `provided_attrs` as a way for a component processor to shuttle params into the the current param resolution for components.  This allows for the writer of a ComponentProcessor to reuse most of the existing ComponentProcessor functionality.  I added an example into the test where a web `request` object is added to component calls as needed from a global `ContextVar`.  The alternative would be to get the value from some sort of application context.

ianjosephwilson:
I made #117 to try to demo multiple integration strategies because it is confusing that this is sort of part of an answer.  I'm going to close this for now.

PR #119, opened 2026-04-28
Middleware API Demo

ianjosephwilson:
This should be stacked on #118 

- ComponentProcessor(s) can now return either `Template` or a 2-tuple of `Template` and a "middleware" providing `object`.
- Only "middleware" currently supported is the ability to set context variable values that affect a component's children.
- Demos purpose of returning a "mystery" object back to the `TemplateProcessor`.


davepeck:
Okay to close at this point?

PR #121, opened 2026-05-06
Move component children template extraction to the parser.

### LINKED ISSUE #120: Refactor component children extraction
There are two things here: 
1. Mark the template to cut it during processing, I guess we could store the strings and slice the interpolations, just cutting it out seems cleaner but slower?
2. ~~Preparse the template in the parser to a tnode tree so it's cached in advance~~
    - Punt this into #134

### LINKED ISSUE #130: Scanning for gt char, `>`, to detect component body fails when gt char is in literal attribute value.
@davepeck saw this issue in #121 but it seems to affect a similar technique in the current processor.

```python
>>> def Link(children: Template, href: str, **attrs) -> Template:
...     return t'<a href={href} {attrs}>{children}</a>'
...     
>>> assert html(t'<{Link} href="/popsicle-spill" title="0 > Temperature"><span>chill out</span></{Link}>') == '<a href="/popsicle-spill" title="0 &gt; Temperature"><span>chill out</span></a\
>'
Traceback (most recent call last):
  File "<python-input-22>", line 1, in <module>
    assert html(t'<{Link} href="/popsicle-spill" title="0 > Temperature"><span>chill out</span></{Link}>') == '<a href="/popsicle-spill" title="0 &gt; Temperature"><span>chill out</span></a>'
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
>>> assert html(t'<{Link} href="/popsicle-spill" title="0 &gt; Temperature"><span>chill out</span></{Link}>') == '<a href="/popsicle-spill" title="0 &gt; Temperature"><span>chill out</span>\
</a>'
>>> html(t'<{Link} href="/popsicle-spill" title="0 > Temperature"><span>chill out</span></{Link}>')
'<a href="/popsicle-spill" title="0 &gt; Temperature"> Temperature&#34;&gt;<span>chill out</span></a>'
```

ianjosephwilson:
Move component children template extraction to the parser.

- TemplateParser: Track the current strings index in addition to the already tracked current interpolations index.
- TemplateParser: Precompute a component's children template into a `TemplateRef` called `children_ref`.
    - Use current strings index and current interpolations index.
    - Use `get_starttag_text` from Python's HTML parser to get the original starttag text.
- TemplateProcessor: Each time we process a component move the current interpolations into `children_ref` to get a current `children` `Template` to pass during invocation.

Fixes #120 
Fixes #130
Punts on #134

### DIFF
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -177,13 +177,11 @@ class TemplateParser(HTMLParser):
             case OpenTComponent(
                 start_i_index=start_i_index,
                 attrs=attrs,
-                children=children,
             ):
                 return TComponent(
                     start_i_index=start_i_index,
                     end_i_index=endtag_i_index,
                     attrs=attrs,
-                    children=tuple(children),
                 )
 
     def validate_end_tag(self, tag: str, open_tag: OpenTag) -> int | None:
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -5,7 +5,7 @@ from string.templatelib import Interpolation, Template
 
 from .htmlspec import VOID_ELEMENTS
 from .placeholders import PlaceholderState
-from .template_utils import combine_template_refs
+from .template_utils import TemplateRef, combine_template_refs
 from .tnodes import (
     TAttribute,
     TComment,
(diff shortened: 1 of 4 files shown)


ianjosephwilson:
Using `get_starttag_text` seems to be a better fit here.  We can actually use the parser itself to determine the end of the starttag rather than trying to scan for ">".

davepeck:
> Using `get_starttag_text` seems to be a better fit here. We can actually use the parser itself to determine the end of the starttag rather than trying to scan for ">".

Agree, this is the way. :-)

This looks great to me; going to merge.

One detail we can address later, @ianjosephwilson : is `OpenTComponent.children` actually used anywhere? Seems like it's dead, and that validation just happens through parsing itself?

Issue #88, opened 2025-12-17
Implement a "fast path" for rendering: no type checks, maybe no `Node`?


davepeck:
Currently:

1. Our processor checks the type of every interpolation on every call to `html()`; if the developer _knows_ that the types of interpolations will never vary for two templates with the same `.strings` tuple, we should offer a fast path that cuts out the extraneous checks.

2. `html()` returns a `Node` tree, even though often all we want is a `str`

This issue asks us to consider and address both.

/cc @ianjosephwilson

pauleveritt:
At very, very long last, I have a first cut at [tdom-svcs](https://pauleveritt.github.io/tdom-svcs/node.html) with a writeup on Node.

The [README](https://github.com/pauleveritt/tdom-svcs/blob/main/README.md) explains that this also includes an implementation of `html(t'<div>hi</div>', context=some_dict)` to avoid prop drilling.

ianjosephwilson:
> When templates produce strings, every tool in the chain must parse HTML to understand the structure. This is slow, error-prone, and loses semantic information.

Just to clarify, I mostly talk about trying to use `Template` instead of `Node`.  When `Template`s are maintained from start to finish the semantic information created in the code is maintained.  When the `Template` is converted to a `Node` tree the semantic information is lost, or at least the separation of "interpolations" from the content itself is lost.  New semantic information might be found, the structure of the dom nodes, but the interpolations are destroyed.   In the same way that generating html strings loses that information.  The `TNode` tree extracts the semantics of the structure AND maintains the interpolations via TComponent, the dynamic TAttribute(s) and the `TemplateRef` structures.  So there are 2 "kinds" of semantics: structure and interpolations.

By using `Template` throughout we can extract the "structure" semantics one-time and then re-use it forever but still keep the "interpolation" semantics as-is and their locations can be re-used between locations.  My intention earlier was how can we let the developer put in MORE semantic information into those interpolations and get more leverage but still not ruin the regular use-cases.  So allow optimization but don't require it.

ianjosephwilson:
> Hey [@ianjosephwilson](https://github.com/ianjosephwilson) up for a call some time? I think what you're saying is fascinating and hope there some clever way to keep both semantics: template semantics and HTML semantics.

Yeah we should schedule something.  Maybe we could sort through some things and then when Dave has some time freed up we all could figure out a way forward.

We do that now but the node structure cannot be fully "realized" without destroying the intermediate structure:
```python
[...]

PR #122, opened 2026-05-06
Runtime checking linter .. "thing"

ianjosephwilson:
Just a poc for linting a template while processing a template, there are a lot of restrictions.


ianjosephwilson:
I think we can just close this for now and I can rebase it if there is a next time.  It was more of a demo.

davepeck:
@ianjosephwilson Yeah, I was sorta contemplating some of the ideas in it, not intending to actually review.

For me, this touches on all the open questions about when and how aggressively we should check types. Obviously, we know we want static checking. We'd like `tdom` to expose really good errors at runtime too, so aggressive type checks might be valuable. And of course we do a decent amount of type-checking for dispatch already. Somehow you'd like all these expensive checks to sit on one side of a cache, if the user is willing to let us assume that a correct check the first time leads to stable types the next time a given template is rendered (possibly with a cheap re-check that falls back if not)?

ianjosephwilson:
The fallback idea sounds feasible.  Something like another cache with the expected interpolation types indexed by `i_index`, sort of like a `TemplateSpec` to parallel `ParamSpec`.  We'll have to check and see how feasible it is to jump from those to something actionable and if we really gained enough to make it worthwhile.  Maybe I'll do some research later and put it into #88 .

PR #131, opened 2026-05-23
Context & Provider

davepeck:
With this PR, I've tried to gather up all the conversations we had at Pycon and build something that will be:

1. [Familiar to React developers](https://react.dev/learn/passing-data-deeply-with-context)
2. Compatible with the broad python web ecosystem (contextvars work well in async/threaded contexts)
3. Flexible enough for advanced scenarios (for instance, integration into Django middleware)

Here's a simple example:

```python
from tdom import create_context

theme = create_context(default="auto")

def Banner() -> Template:
    return t"<h1 class={theme.get()}>Hello!</h1>"

def App() -> Template:
    return t'<{theme.Provider} value="dark"><{Banner}/></{theme.Provider}>'
```

`theme` is a module-level global that defines our context. All contexts created via `create_context()` _must_ be module-level. They must also have a default value.

Here, when `html(t"<{App} />")` is called, the `Banner` component will get a `dark` value for the `theme`. But if you called `theme.get()` in `App()`, you'd get the default value, `auto`. 

`theme` is of type `tdom.Context[T]`. The `Context` type has three things:

1. a `get()` method
2. a `Provider` component that you can nest in your t-strings, and that sets the value for the sub-tree
3. a `provide()` context manager, for the Python equivalent of the `Provider` component in imperative code.

(removed earlier broken example, see comments below)

```python
def App() -> Template:
    return t'<{Banner}/>'
    
with theme.provide("dark"):
    return html(t'<{App} />')
```

Under the hood, `Context[T]` builds on python's `ContextVar[T]`. When you call `create_context()`, we create a context variable for you. For more advanced cases where you want direct control over the `ContextVar` itself, you can create one directly and then call `Context(my_cv)` to get a `Context`. This allows you more control, for instance if you'd like to set the `request` context in Django middleware. (This also allows you to _not_ set a default; if a template attempts to read a value for an unset context, an exception will be raised.)

Component processing now returns `Template | ScopedTemplate`, where the scope indicates which context variable to set for the child hierarchy. This is just a variant on Ian's earlier `Template | tuple[Template, SomethingForContext]` suggestion. Unless devs are actively using our component extension point, this is an implementation detail they can be completely unaware of.

Give it a go; it's felt pretty good in my early usage!


davepeck:
I haven't written any docs yet; figure this will get balled up in the larger effort. 

So I don't forget, when we _do_ write docs, I think it's important to set the stage for when `Context` is useful and when it's probably not the right solution. 

To name one example that came up when I was working on `django-tdom`: `Context` isn't the right tool for passing the results of a database query into your template from within a view function in Django; `Context` is really for cross-cutting ambient state: things like theme, locale, current user, request, feature flags, etc. Prop drilling is always the place to start; beyond that, there are plenty of available approaches to getting the right data into your templates that are a little fancier than prop drilling but not as global-heavyweight as `Context`.

tdom/context.py
```diff
@@ -0,0 +1,200 @@
...
+    def App() -> Template:
+        return t'<{theme.Provider} value="dark"><{Banner}/></{theme.Provider}>'
+
+Inside the Provider's subtree, `theme.get()` returns `"dark"`. Outside,
+it returns whatever the surrounding scope had: the default, or
+whatever another outer Provider set. Providers nest as you'd expect: the
+innermost one wins inside its own children, and everything resets
+cleanly on the way out (even if an exception is raised).
+
+One important caveat: both `make_provider` and `create_context` must be
```
ianjosephwilson: I think only `create_context` needs to be top-level.

tdom/context.py
```diff
@@ -0,0 +1,200 @@
...
+        return t'<{ThemeProvider} value="dark"><{Banner}/></{ThemeProvider}>'
+
+    Consumers inside the subtree see `value` via `theme.get()`.
+    """
+
+    def Provider(value: T, children: Template) -> ScopedTemplate:
+        return ScopedTemplate(template=children, scope=Scope(cv=cv, value=value))
+
+    Provider.__name__ = f"{cv.name}Provider"
+    Provider.__qualname__ = Provider.__name__
```
ianjosephwilson, line 124: What is the purpose of `__qualname__` here?  I actually had not heard of it.  Even after reading about it I still don't understand it.  Shouldn't we want to use it to determine that this component was created dynamically inside this function ?

tdom/context.py
```diff
@@ -0,0 +1,200 @@
...
+        exception), the previous value is restored:
+
+            with theme.provide("dark"):
+                assert theme.get() == "dark"
+            # theme.get() returns whatever it was before
+        """
+        return Scope(cv=self._cv, value=value).activate()
+
+
+def create_context[T](
```
ianjosephwilson, line 173: I almost feel like maybe the Pythonic implementation should go in as one PR and the react shim should maybe go in as another PR and/or maybe in another module?

tdom/context.py
```diff
@@ -0,0 +1,200 @@
...
+    You can also use the returned `Context` directly, without embedding
+    a `Provider` inovcation in your t-string:
+
+        theme = create_context(default="auto")
+
+        def Banner() -> Template:
+            return t"<h1 class={theme.get()}>...</h1>"
+
+        def App() -> Template:
+            with theme.provide("dark"):
```
ianjosephwilson: I don't think calling the context manager like this will work because the template is returned before the processing starts.

tdom/processor.py
```diff
@@ -594,6 +596,13 @@ def process(
         The primary purpose of this style is to support
         using a `dataclass` with `def __call__(self) -> Template` as a
         component.
+
+        3. Context-provider component
+
+        Calling `component_callable` returns a `ScopedTemplate`. The
+        template processor recognizes it, activates the scope, processes
+        the inner template, and resets. Created via
+        `tdom.make_provider(cv)` / `tdom.create_context(...)`.
```
ianjosephwilson: My initial feeling is that this should just be another return form of both the other "styles" and not a third style.

tdom/processor.py
ianjosephwilson: I think we need the `ScopedTemplate` check here as well.  So that factory components can also return a `ScopedTemplate`.  Although I saw the reasoning above and now I see why its not here.  In some ways this might be easier to explain to new users but I think it actually makes the api more confusing.

tdom/processor.py
```diff
@@ -844,6 +853,9 @@ def _process_component(
         result_t = self.component_processor_api.process(
             template, last_ctx, component_callable, attrs, children_template
         )
+        if isinstance(result_t, ScopedTemplate):
+            with result_t.scope.activate():
+                return self._process_template(result_t.template, last_ctx)
```
ianjosephwilson, line 857: I think this part looks 10 times better than my `tuple` API.

ianjosephwilson (commented):
Overall I really like `ScopedTemplate` in place of `tuple[Template, object]`.  It conveys the meaning much better and gets us out of the `context` all-the-things problem.  The `instance` checks also look really cohesive in the processing.

I made some comments about keeping the api symmetric between factory/`class` and `function` components.

I was hoping we could find a better way to express accessing the scope within components.  Maybe that could be in another PR and maybe even provided via extension, ie. `def Consumer(theme: Annotated[str, get_current_theme]) -> Template: ...`.

Mixing react familiarity with the context interaction feels like too many cake layers at once.  Maybe this could be cut back to`ScopedTemplate` and `Scope`?  That could get to `scopes.py` which keeps everything distinct from `context`.  Furthermore a generic extension point could be provided with something like this:
```python
#... 
@dataclass(frozen=True, slots=True)
class ScopeProvider[T]:
    children: Template
    cvar: ContextVar[T]
    cvalue: T

    def __call__(self) -> ScopedTemplate:
        return ScopedTemplate(
            template=self.children, scope=Scope(cv=self.cvar, value=self.cvalue)
        )
```

This type of stuff feels like it should be set up by someone more advanced and then used by tdom users.  Or just setup outside the regular plug and chug component usage. Something like this:

```python
from contextvars import ContextVar
from string.templatelib import Template

from .scopes import ScopeProvider

ThemeCtx: ContextVar[str] = ContextVar("ThemeCtx", default="auto")


def get_current_theme() -> str:
    """
    Get the theme when provided, otherwise return the default.
    """
    return ThemeCtx.get()


def ThemeProvider(children: Template, theme: str) -> ScopeProvider[str]:
    """
    Component which provides a specific theme to the given children.
    """
    return ScopeProvider(children=children, cvar=ThemeCtx, cvalue=theme)
```

tdom/context.py
davepeck: Yeah, sorry. I botched this example when I wrote up my PR. 

`.provide()` is only useful at a render boundary, so *this* is what I should have written:

```python
def App() -> Template:
	return t'<{Banner} />'

with theme.provide("dark"):
	return html(t'<{App} />)
```

tdom/context.py
davepeck: Agree, good catch.

tdom/context.py
davepeck: Primarily it's useful for tracebacks and debugging, since this is dynamically created. Maybe I should comment this.

tdom/processor.py
davepeck: Agree, bad docs on my part. Fixed.

tdom/processor.py
davepeck: Agree; I missed this. Fixed.

tdom/processor.py
davepeck: I was thinking 1.001 times better :-P

tdom/context.py
davepeck: I think my instinct is that this is a tight enough unit to keep as one PR.

davepeck:
Thanks, @ianjosephwilson for the review! I've addressed a bunch of the details, and fixed my bad example with `.provide()`.

Some other comments:

> Mixing react familiarity with the context interaction feels like too many cake layers at once.

I definitely get this, but I think the answer is probably good documentation rather than trying to remove the `.Provider` layer. The intent is that most users will use `create_context(...)` and `context.Provider`; middleware might want to bring its own `cv` and/or use `.provide()`; advanced stuff might want to work with `Scope`/`ScopedTemplate` directly.

> I was hoping we could find a better way to express accessing the scope within components. Maybe that could be in another PR and maybe even provided via extension, ie. `def Consumer(theme: Annotated[str, get_current_theme]) -> Template: ....`

I kinda sit on the fence here. `.get()` feels pretty straightforward and explicit to me.

That said, I dunno! We should probably explore this dependency injection style approach in a follow-on PR and see how it feels. (There's also a trivial `@consumer` decorator approach that could be wrapped around component functions to make their consumption explicit, I think?)

pauleveritt:
I'm not sure if the `Provider` approach is still canonical in React-land. I had the impression people had moved on, and some still use it, while others use other approaches?

If people want alternative approaches, perhaps integrated into existing lifecycles (e.g. Starlette/FastAPI), will that be ok? Can I still use the approach that I showed you, to let me keep using `svcs` and DI?

davepeck:
I _think_ this approach leaves sufficient escape valves for you to do what you want without running into trouble with `svcs`. But would want to explore, to be sure.

pauleveritt:
@davepeck I think I'd do a variation of what I was doing before with a custom processor, right?

davepeck:
> @davepeck I think I'd do a variation of what I was doing before with a custom processor, right?

I think so, yes! But can you point me to code? I'll take a look and see if this PR helps/hinders/is neutral...

pauleveritt:
Sure, thanks for looking.

1. [My custom component processor](https://github.com/pauleveritt/tdom-svcs/blob/main/src/tdom_svcs/processor.py#L333) is passed the `container` to store on the instance. The `process()` method, when a container is present, does some mumbo-jumbo for sniffing at signatures. 

2. How is this processor created? With a [small _make_processor() helper](https://github.com/pauleveritt/tdom-svcs/blob/main/src/tdom_svcs/processor.py#L404).

3. What calls that? The top of your call chain can do `html(t"<div>Hi</div>", container=request_container)`. This is a custom `html()` function in my package. It [calls _make_processor](https://github.com/pauleveritt/tdom-svcs/blob/main/src/tdom_svcs/processor.py#L435) if the processor isn't already in the container.

pauleveritt:
I realize this stuff is obscure if you haven't looked at it. If you'd like, I can record a 5 min walkthrough.

ianjosephwilson:
Yeah I guess I would think that frameworks would probably not use `provide` and just wrap `html()` in a `with` call using a `ContextVar`.  Or at least we should keep supporting that.  I would think something like this:

|User Type|Examples|Crosscutting State Delivery|Implementor|
|---|---|---|---|
|**Framework**|flask , django, starlette, pyramid|wrap `html()` with `with framework_context_var.set()`|framework + `tdom` integration library|
|**Library**|theme libraries, 3rd party auth libraries, libraries that extend frameworks themselves | `ScopedTemplate` method (this PR) | library itself via `tdom` helpers |
|Normal App / Website | Blog/website with some dynamic content| uses **Library** and/or **Framework** | someone else |
| Large App | Website with multiple responsibilities: admin, content, ecommerce, etc. | users **Library**, **Framework** and/or `ScopedTemplate` method (this PR) | large maintainer AND/OR others |

davepeck:
@pauleveritt Thanks for the links.

After looking it over, I don't think this PR alters your story or breaks what you're doing. 

It does require that you update the [return type annotation](https://github.com/pauleveritt/tdom-svcs/blob/main/src/tdom_svcs/processor.py#L341) for your `process(...)` method to `-> Template | ScopedTemplate`. Other than that, no changes needed.

The two things from `tdom-svcs` I keep thinking about for `tdom`: (1) whether the per-render parameter to `html()` (`container`) has value, and (2) whether something more DI-like (your filling-in for `Inject[T]`, or Ian's `Annotated[...]`) belongs here. The good news is our current extension point doesn't seem to preclude creating such a thing.

davepeck:
> Yeah I guess I would think that frameworks would probably not use `provide` and just wrap `html()` in a `with` call using a `ContextVar`.  Or at least we should keep supporting that. 

The Scope type is explicitly intended for this; it's just a tiny context-manager friendly wrapper around a ContextVar. (ContextVar.set() isn't itself a context manager.)

I think the thing that's missing here is that web developers building their own apps will probably reach for Provider. The intent is that's the most common approach even when you're building on (say) Django with tdom-django. The underlying integrations can use the right lower-level machinery as needed. 

But I need to build both the app and the integration and feel it out before I'm sold on all this.

ianjosephwilson:
It is weird but the `Token` is a context manager as I understand it so that you can use `cvar.set()` that way.

https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar.set

> For convenience, the token object can be used as a context manager to avoid calling ContextVar.reset() manually:

I have been trying it like this:

https://github.com/ianjosephwilson/flask_tdom/blob/main/src/flask_tdom/__init__.py#L35

This only works for the top-most level though (unlike this PR).

EDITED:

The reason I use that `ContextVarSetter` in a lot of example was that I was setting multiple context vars at once all in the same context manager.

pauleveritt:
Thanks @davepeck for looking at it. I'll keep plowing ahead and if your work lands, I'll make the update you suggested.

davepeck:
Oh, that _is_ weird. Somehow I'd missed that about ContextVar tokens since, like, forever. Thanks.

So, okay, maybe break this into (a) Scope (still useful, in that it couples a CV and a value) and ScopedTemplate, to merge today, and (b) Context and Provider, to a separate draft PR pending how they shake out in our real apps and integrations?

ianjosephwilson:
@davepeck Sounds good to me.

> Oh, that's _is_ weird. Somehow I'd missed that about ContextVar tokens since, like, forever. Thanks.
> 
> So, okay, maybe break this into (a) Scope (still useful, in that it couples a CV and a value) and ScopedTemplate, to merge today, 

This foundation seems solid.  `Scope` almost seems extra now but later on the concept could stay the same but could be extended to involve multiple cvs at once (for example).

> and (b) Context and Provider, to a separate draft PR pending how they shake out in our real apps and integrations?

Yeah I think there is something here but since the foundation (a) is set things won't get out of hand if we don't build this part right now.  We can experiment for a bit and see what makes the most sense.  Also famous last words, we can also make sure the foundation is working as we intended first as well.

davepeck:
> we can also make sure the foundation is working as we intended first as well

Let's not get _too_ crazy...

davepeck:
To draft, after deciding to split this into two. See #135

davepeck:
Closing for now.

Issue #129, opened 2026-05-22
Dangling quote variation fails silently


ianjosephwilson:
This is a variation of the dangling quote problem in #99 but a little more serious because no errors occur.  Even whitespace would be collected into a `TText`.  We need to check if data was fed in but no handlers were called.  `starttag` is never reached because the parser is still waiting for the rest of the attribute...

```python
assert TemplateParser.parse(t'<div x=">') == TFragment(children=())
assert html(t'<div x="') == '' # no error
assert html(t'<div x=">') == '' # no error
```

ianjosephwilson:
Tracking the failure to move the `pos` might be a starting point:

```python
>>> parser = TemplateParser()
>>> parser.getpos()
(1, 0)
>>> parser.feed_template(t'<div x=">')
>>> parser.getpos()
(1, 0)
>>> parser.close()
>>> parser.get_tnode()
TFragment(children=())
```

With a "complete" unclosed tag the `pos` is moved forward as the start tag is parsed:

```python
>>> parser = TemplateParser()
>>> parser.getpos()
(1, 0)
>>> parser.feed_template(t'<div>')
>>> parser.getpos()
(1, 5)
>>> parser.close()
Traceback (most recent call last):
  File "<python-input-62>", line 1, in <module>
    parser.close()
    ~~~~~~~~~~~~^^
  File "/home/ian/workspace/pep750/main/tdom/tdom/parser.py", line 291, in close
    raise ValueError("Invalid HTML structure: unclosed tags remain.")
ValueError: Invalid HTML structure: unclosed tags remain.
```

Using this internal variable appears to work ... probably not a good idea to depend on it though but we can use it to determine if our source tracking is working or not.

```diff
diff --git a/tdom/parser.py b/tdom/parser.py
index 5ac501b..bc40351 100644
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -291,6 +291,8 @@ class TemplateParser(HTMLParser):
             raise ValueError("Invalid HTML structure: unclosed tags remain.")
         if not self.placeholders.is_empty:
             raise ValueError("Some placeholders were never resolved.")
+        if self.rawdata:
+            raise ValueError("Some text was not processed.")
         super().close()
 
     # ------------------------------------------
```

ianjosephwilson:
Fixed in #133

Issue #99, opened 2026-01-11
Dangling attribute quote raises the wrong error message


ianjosephwilson:
```python
# missing closing "
node = html(t'<p class="{classes}></p>')
```

This fails saying the placeholder is not found which is not very helpful.  I'm not sure if we are failing before a proper message or we need to detect it ourselves.  Just putting this here so it gets dealt with at some point.
```
    def close(self) -> None:
        if self.stack:
            raise ValueError("Invalid HTML structure: unclosed tags remain.")
        if not self.placeholders.is_empty:
>           raise ValueError("Some placeholders were never resolved.")
E           ValueError: Some placeholders were never resolved.

tdom/parser.py:293: ValueError
```

ianjosephwilson:
Another confusing situation where we are missing the space before the self-closing component tag:

```python
from string.templatelib import Template

from tdom import html


def AuthStatus(logged_in: bool = False) -> Template:
    msg = 'In' if logged_in else 'Out'
    return t'<div>{msg}</div>'


test_t = t'<{AuthStatus} logged_in={True}/>'

try:
    print (html(test_t))
except ValueError as e:
    print (e)


test_t = t'<div><{AuthStatus} logged_in={True}/></div>'

try:
    print (html(test_t))
except ValueError as e:
    print (e)


test_t = t'<div><{AuthStatus} logged_in={True} /></div>'

try:
    print (html(test_t))
except ValueError as e:
    print (e)
```

```text
Invalid HTML structure: unclosed tags remain.
Mismatched closing tag </div> for component starting at AuthStatus.
<div><div>In</div></div>
```

ianjosephwilson:
Moved the second issue into #132

ianjosephwilson:
Fixed in #133

PR #133, opened 2026-05-24
Temp solution to incomplete parsing issue

ianjosephwilson:
I think we just have to depend on `rawdata` for now until we find a better solution.  I tried to wrap it in a method so we know why we are using it.

Fix for initial issue in #99 
Fix for #129


davepeck:
Sigh. Yes, baby steps I guess. :-)

PR #135, opened 2026-05-25
ScopedTemplate

davepeck:
After discussion on #131 we've decided to cleave off `Scope` and `ScopedTemplate` into its own PR. `ScopedTemplate` is the minimal surface area we need to allow others to build context-style features on top of `tdom`. 

We'll revisit `Context` and `Provider` from 131 if they prove useful in our own integration/app dev work.


ianjosephwilson:
This looks good to me.

Issue #132, opened 2026-05-24
unquoted attribute at end of self-closing tag consumes trailing slash


ianjosephwilson:
Confusing situation where we are missing the space before the self-closing component tag:

```python
from string.templatelib import Template

from tdom import html


def AuthStatus(logged_in: bool = False) -> Template:
    msg = 'In' if logged_in else 'Out'
    return t'<div>{msg}</div>'


test_t = t'<{AuthStatus} logged_in={True}/>'

try:
    print (html(test_t))
except ValueError as e:
    print (e)


test_t = t'<div><{AuthStatus} logged_in={True}/></div>'

try:
    print (html(test_t))
except ValueError as e:
    print (e)


test_t = t'<div><{AuthStatus} logged_in={True} /></div>'

try:
    print (html(test_t))
except ValueError as e:
    print (e)
```

```text
Invalid HTML structure: unclosed tags remain.
Mismatched closing tag </div> for component starting at AuthStatus.
<div><div>In</div></div>
```

ianjosephwilson:
The interpolation just makes this seem like it should work but this same issue comes up with a regular unquoted attribute as well. 

```python
assert html(t'<div x=a />') == '<div x="a"></div>'
# raises ValueError: Invalid HTML structure: unclosed tags remain.
html(t'<div x=a/>') 
```

davepeck:
Ugh, yup.

ianjosephwilson:
I kind of think the best we could do here is just try to introspect things during the "unclosed tags remain" error and if it looks like there is a dangling "/" in the last attribute of that tag then we'd put in a different error message, ie. `'If you meant to self-close the tag {tag} then you need a space before the "/".'`.  Seems "not great" but trying to outguess the input seems worse.  We could also just leave it as-is, ie. do nothing, and see how often it comes up.

That being said I think we might reconsider how we handle self closing tags because it is similar to react but it seems non-standard and might cause problems later on.  Maybe we should only allow void elements, components and foreign elements to be self-closing.  Then we could consider components to be "foreign elements".  Maybe we were already doing that and I'm out of the loop but we would have guidance and could maybe add/change parsers later on.

[syntax-start-tag](https://html.spec.whatwg.org/multipage/syntax.html#syntax-start-tag)

[...]

PR #136, opened 2026-06-01
Raise an error if xhtml-style "/" is ambiguous.

ianjosephwilson:
If we can't tell if "/" should be part of an attribute's value versus a self-closing tag then raise an error.  Potential solution to #132


ianjosephwilson:
Closing for now.

PR #137, opened 2026-06-01
Self-close if xhtml-style "/>" is ambiguous. 

ianjosephwilson:
Don't let unquoted attribute value's consume the "/" if it seems like the user meant to self-close a tag.  Potential solution to #132.  Alternative to #136.


ianjosephwilson:
This is stacked on #136.

ianjosephwilson:
Closing for now.

PR #138, opened 2026-06-01
Only allow components to self-close and correct simple "/>" ambiguity.

ianjosephwilson:
This is a 3rd option forward for #132  

I still need to research if there is a simpler way to deal with foreign elements like svg/mathml.


ianjosephwilson:
Closing for now.

PR #140, opened 2026-06-12
Improve error message for unclosed components whose attribute value might have consumed the slash.

ianjosephwilson:
Try to give more detailed error messages when a component's unquoted attribute consumes the "/" AND an error occurs where it is not closed.

This is the first step in trying to get through #132 

Examples that fail but were meant to probably self-close the component: 
    - `t"<{Comp} title={title}/>"`
    - `t"<div><{Comp} title={title}/></div>"` 
    - `t"<{Comp2}><{Comp1} title={title}/></{Comp2}>""`

The last example requires inspecting the interpolations to resolve that `Comp1 != Comp2`.

This brings up a past issue we tried to skip over where this would parse but fail processing:
- `t"<{Comp1}></{Comp2}>"`

### DIFF
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -189,11 +189,9 @@ class TemplateParser(HTMLParser):
         # @NOTE: This must be called when the tag is handled since it is
         # populated based on the most recently finished start tag. Otherwise
         # the value will be out of sync.
-        starttag_text = self.get_starttag_text()
-        if starttag_text is None:
-            raise AssertionError(
-                f"Expected startag_text to be set when parsing component at {i_index}."
-            )
+        starttag_text = self.always_get_starttag_text(
+            f"Expected startag_text to be set when parsing component at {i_index}."
+        )
 
         tattrs = self.make_tattrs(attrs)
 
@@ -370,6 +368,19 @@ class TemplateParser(HTMLParser):
                 # component callable that matches the start tag. We do not check
                 # any of this in the parser, instead relying on higher layers.
                 return tag_ref.i_indexes[0]
+
+    def always_get_starttag_text(
+        self, msg: str = "Expecting starttag text to be set."
+    ) -> str:
+        """
+        Wrap get_starttag_text and just raise if None is returned.
+
+        Do this so we don't guard for `None` everywhere.
+        """
+        starttag_text = self.get_starttag_text()
+        if starttag_text is None:
+            raise AssertionError(msg)
+        return starttag_text
 
     # ------------------------------------------
     # HTMLParser tag callbacks
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -93,13 +93,6 @@ class SourceTracker:
                 f"Interpolation indices exceed bounds: {index1} {index2}: [0...{last_index}]"
             )
 
-    def expressions_match(self, i_index1: int, i_index2: int) -> bool:
-        self._check_indices(i_index1, i_index2)
-        return (
-            self.interpolations[i_index1].expression
-            == self.interpolations[i_index2].expression
-        )
-
     def values_match(self, i_index1: int, i_index2: int) -> bool:
         self._check_indices(i_index1, i_index2)
         return (
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -471,6 +471,33 @@ class TemplateParser(HTMLParser):
         self.placeholders = PlaceholderState()
         self.source = None
 
+    def has_ambiguous_forward_slash(self, open_tag: OpenTag) -> bool:
+        """
+        Detect when an unquoted attribute value consumes a trailing "/" that
+        *might* have been meant to attempt to self-close a tag, ie. "/>".
+
+        This can come up with literal values or values with interpolations.
+
+        Such as "<div title=test/>" or "<{Component} title=test/>".
+
+        Or more often "<{Component} title={title}/>" which should be corrected
+        with "<{Component} title={title} />".
+        """
+        if isinstance(open_tag, (OpenTElement, OpenTComponent)):
+            return (
+                # has attributes
+                len(open_tag.raw_attrs) > 0
+                # last attr not bare attribute
+                and open_tag.raw_attrs[-1][1] is not None
+                # last char of last attr is "/"
+                and open_tag.raw_attrs[-1][1][-1] == "/"
+                # parsed starttag ends with "/>"
+                and open_tag.starttag_text.endswith("/>")
+                # if parsed as startend then its not ambiguous
+                and not open_tag.startend
+            )
+        return False
+
     def close(self) -> None:
         if self.waiting_for_data():
             # We apply heuristics here to try to guess why the parser didn't finish.
@@ -483,7 +510,24 @@ class TemplateParser(HTMLParser):
                     "Parser expects more data, is the template valid html?"
                 )
         if self.stack:
-            raise ValueError("Invalid HTML structure: unclosed tags remain.")
+            e = ValueError("Invalid HTML structure: unclosed tags remain.")
+            # Check for tags that might have meant to self-close but whose
+            # unquoted last attribute value consumed a "/", ie. <div id=app/>.
+            parent = self.stack[-1]
+            if isinstance(parent, (OpenTElement, OpenTComponent)):
+                if isinstance(parent, OpenTElement):
+                    starttag = parent.tag
+                elif isinstance(parent, OpenTComponent):
+                    starttag = (
+                        f"{{{self.get_source().format_starttag(parent.start_i_index)}}}"
+                    )
+                if self.has_ambiguous_forward_slash(parent):
+                    e.add_note(
+                        f'Did you mean to quote the last attribute or put a space before "/>" for "<{starttag} .../>"?'
+                    )
+                else:
+                    e.add_note(f"Most recently unclosed tag is <{starttag} ...>")
+            raise e
         if not self.placeholders.is_empty:
             raise ValueError("Some placeholders were never resolved.")
         super().close()
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -26,8 +26,8 @@ type HTMLAttributesDict = dict[str, str | None]
 
 
 @dataclass(frozen=True)
-class ParseInfo:
-    "Track parse info for error reporting."
+class TagOpening:
+    "Track parse info at the opening of a tag for error reporting."
 
     starttag_text: str
     " Entire starttag as parsed, includes placeholders, . "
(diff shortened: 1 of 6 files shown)


tdom/parser_test.py
```diff
@@ -425,15 +425,17 @@ def Component():
...
 
     def Component2():
         pass
 
-    node = TemplateParser.parse(t"<{Component1}></{Component2}>")
-    assert node == TComponent(start_i_index=0, end_i_index=1)
+    with pytest.raises(
+        TypeError, match="Component start and end tags must contain the same callable."
+    ):
+        _ = TemplateParser.parse(t"<{Component1}></{Component2}>")
```
davepeck: Love that we're revisiting this...

tdom/placeholders.py
```diff
@@ -61,6 +61,9 @@ class PlaceholderState:
     config: PlaceholderConfig = field(default_factory=make_placeholder_config)
     """Collection of currently 'known and active' placeholder indexes."""
 
+    def copy(self):
```
davepeck, line 64: Is this used in 140, or hanging out because of 141?

tdom/parser.py
```diff
@@ -27,6 +27,12 @@
 
 @dataclass
 class OpenTElement:
+    starttag_text: str
```
davepeck, line 39: I like these additions. 'Used for debugging' seems not quite right, since they're used in `has_ambiguous_forward_slash` itself.

tdom/parser.py
```diff
@@ -359,18 +400,73 @@ def validate_end_tag(self, tag: str, open_tag: OpenTag) -> int | None:
...
+                        )
+                    raise e
                 if not tag_ref.is_singleton:
                     raise ValueError(
                         "Component end tags must have exactly one interpolation."
                     )
-                # HERE BE DRAGONS: the interpolation at end_i_index shuld be a
-                # component callable that matches the start tag. We do not check
-                # any of this in the parser, instead relying on higher layers.
+                if not source.expressions_match(
```
davepeck: ...but there's a catch, which is that "cache only on `.strings`" is problematic now that we're looking at interpolation values in the parser.

```python
def C1():
    pass

def C2():
    pass

proxy = CachedTemplateParserProxy()
proxy.to_tnode(t"<{C1}></{C1}>")  # works as expected
proxy.to_tnode(t"<{C1}></{C2}>")  # doesn't raise because we hit the cache
```

I'm not sure exactly how to think about this. I know _some_ check is necessary to handle your `t"<{C2}><{C1} attr=value/></{C2}></{C2}>"` case. But I'm reluctant to change our caching story.

I wonder if we could compute `has_ambiguous_slash` in the parser and stash it on the `TComponent`, leaving it up to the processor to surface a good error message.

tdom/parser.py
```diff
@@ -449,7 +545,21 @@ def close(self) -> None:
                     "Parser expects more data, is the template valid html?"
                 )
         if self.stack:
-            raise ValueError("Invalid HTML structure: unclosed tags remain.")
+            e = ValueError("Invalid HTML structure: unclosed tags remain.")
+            # Check for tags that might have meant to self-close but whose
+            # unquoted last attribute value consumed a "/", ie. <div id=app/>.
+            parent = self.stack[-1]
+            # @TODO: We need to determine which tags this might apply to, this only applies to components.
```
davepeck: Yeah this PR sequences before the work that would answer this. But it does feel like where we land is: anything self closable (components here; in a future PR, void tags; even further, foreign tags once the parser becomes smart about such things) gets one message; everyone else gets the other.

davepeck (commented):
I think we're close, but my instinct is the actual value check still needs to live in the processor (alas). We'd have to pass the result of `has_ambiguous_slash()` through our component instances so the processor can surface a sensible error message. This lets us keep our simple `.strings` cache for parsing. (And later, when 141 lands, `.strings` + a namespace, I guess?)

I dunno, what do you think about that approach?

tdom/parser.py
davepeck: (Random aside: is the expression check before the value check a perf thing? In the common case, it seems good. But I suppose we can craft identical expressions that may not be equivalent, like `t"<{(lambda: 0)}></{(lambda: 0)}>"`)

tdom/parser.py
ianjosephwilson: But `has_ambiguous_forward_slash` is only used when we are trying to make the error message more useful.  I don't think we use it directly, if that makes sense.  I did "clump" them in #141 with a later [commit](https://github.com/t-strings/tdom/pull/141/changes/84a1711abebc1730ddd00fddd741c37e4aa36455) .  Kind of hard to keep re-arranging these but I think it is nicer when they get pulled out together.  They aren't meant to be used "in the moment" of parsing but rather to look info back up when something goes wrong.  At least so far.

tdom/parser.py
ianjosephwilson: I guess, in this case, I am ok NOT involving this in the caching story.  I think it is okay to say that the open tag value and the closing tag value for component tags need to have the same value if you want it parsed correctly ( or always?).  After the "first" time whenever that is as there might be a startup stampede then we'll just assume you passed in the right values otherwise the behavior is "unpredictable".  The only part I don't like is digging in the interpolation values "in general".  I think if for example you wanted to pre-parse the templates then you could pass dummy values that follow that rule and it should still work.  Otherwise we end up with essentially this situation `t'<{C1}><{C2}></*><{C3}></*></*>'`.  I feel like confusing error messages would be worse but this also seems like something we could potentially remove later if it was a problem or even make configurable?

An alternative which might be possible is that when the inevitable failure occurs then we backtrack and try to rerun through the tags and decipher which component was incorrectly closed with a different component for whatever reason, the common case probably being this: `t'<{C1}><{C2} title=/></{C1}>'`, where `</{C1}>` closes `<{C2} title=/>` because the `/` was consumed as part of `title`.  Then the interpolation introspection only occurs during errors.  We could try to make that work?

tdom/parser.py
ianjosephwilson: Yeah I think actually void tags wouldn't get this message because they don't go on the stack but yes we'd probably add foreign tags.  

Just to clarify: This error is happening regardless of the slash but we try to make "Did you mean ..." help message for a potentially hard to debug case but the "unclosed tags remain" is always happening.  They could have *meant* to use the slash as an attribute value but ... just forgot to close the component (or foreign tag).  Ie. `t'<{Comp1} path=/about/>'` instead of `t'<{Comp1} path="/about/" />'`

tdom/parser_test.py
ianjosephwilson: Haha yeah, clearly it is one of the project's hard to scratch itches.  Anytime a large involved change comes along ... so does it.

tdom/placeholders.py
ianjosephwilson: I think maybe this got left in when I tried to split these changes out from #141.  I could probably take it out, do you think you'll have a chance to look at #141 too ?

ianjosephwilson:
> I think we're close, but my instinct is the actual value check still needs to live in the processor (alas). We'd have to pass the result of `has_ambiguous_slash()` through our component instances so the processor can surface a sensible error message. This lets us keep our simple `.strings` cache for parsing. (And later, when 141 lands, `.strings` + a namespace, I guess?)
> 
> I dunno, what do you think about that approach?

I think I confused things by trying to change that test but the original problem actually was trying to make a parser error more helpful for something that was going to fail anyways.  I found a work-around that might work but:
- I misunderstood the original error (although there is a "family" of these types of errors)
- I think I misunderstood your concern about the cache.

So better wait on this and I'll try to put something up again tomorrow.

tdom/parser.py
```diff
@@ -449,7 +576,42 @@ def close(self) -> None:
...
+                        starttag = source.format_starttag(comp.start_i_index)
+                        endtag = source.format_endtag(comp.end_i_index)
+                        e.add_note(
+                            f"Component start tag, <{{{starttag}}}>, and end tag, </{{{endtag}}}>, have values that do not match."
+                        )
+                        if self.has_ambiguous_forward_slash(closed_tag):
+                            e.add_note(
+                                f'Did you mean to quote the last attribute or put a space before "/>" for "<{{{starttag}}} .../>"?'
+                            )
+            raise e
```
ianjosephwilson, line 686: @davepeck This block now performs the component start/end values check but where an exception is already going to happen (we closed the parser but tags are still open on the stack).  So the parsing would have already failed.  I think that should still keep our cache solely dependent on the shape because the values check does not affect what succeeds it is only used to provided better error reporting.  Does that make sense?

tdom/placeholders.py
ianjosephwilson: I ended up using this after all to inspect the end tags.

tdom/parser.py
```diff
@@ -95,12 +122,17 @@ def format_starttag(self, i_index: int) -> str:
...
+        return self.get_expression(i_index, fallback_prefix="component-endtag")
+
 
 class TemplateParser(HTMLParser):
     root: OpenTFragment
     stack: list[OpenTag]
     placeholders: PlaceholderState
     source: SourceTracker | None
+    tmap: dict[TComponent | TElement | TFragment, OpenTag]
+    " Map from completed tnodes back to their opentag for error reporting. "
```
ianjosephwilson, line 147: I added this mapping from the completed tnodes back to the opentags BUT it doesn't include the component bodies that we discard because there are no tnodes made.  This already seems like an issue when we try to report errors based on parse issues deeper inside a component's children.  I'm not sure how to handle that yet.   I still think not trying to preload the cache with the `children` is right but we might need to keep some bookkeeping about what was parsed.  There are going to be some growing pains as this error code balloons up and we try to keep refactoring it back down to sensible levels.  We might need an additional helper class to try to isolate some of that from the primary parsing code.

ianjosephwilson:
I put this on hold to try to prototype pushing the info we need for errors into the processor while juggling them during half-parsed error handling.  I think using the "parser position" from `.getpos()` seems to work as a key and we can keep "open" tag source info during parsing like we do with "open" tags.  Then when the tags are closed we can store it in a table, `sinfo_table`, to both lookup during parsing but also pass on to processing via probably a new `TTree` wrapper structure.  This seemed like a better stopping place to try to get the original errors around the "/>" in.

davepeck:
@ianjosephwilson I know #145 stacks on top but it also seems to _include_ the key pieces of this PR? At this point, it seems to make sense to close this PR? 

Alternatively, I'm happy to merge this PR as-is to `main` now if you'd prefer to stage it that way. It only targets the `<{Component} ... >` cases, but it's still an incremental improvement. That might make sense given that #145 is still a draft and key parts (position translation code, error chaining) are going to need time to bake.

ianjosephwilson:
Closing it sounds fine.  I have been struggling with what is the best way to get feedback without going too far and pushing in a gigantic PR vs spamming PRs.  It sounds like things are beginning to align and I am going to make even more improvements to #145 anyways.

Issue #146, opened 2026-07-08
Support spreading any Mapping as attributes


gazpachoking:
I'm using a utility which was made for other html libraries to be able to [build datastar attributes](https://github.com/starfederation/datastar-python#attribute-generation-helper). Seems it isn't compatible with tdom right now because it returns a `Mapping` rather than a strict `dict`. Can/should tdom support using any `Mapping` type to set element attributes?

https://github.com/t-strings/tdom/blob/5608843ffc63ad2c7fb7283c2c67065fe7166ea9/tdom/processor.py#L145-L146

Ideally we could:
```python
from datastar_py import attribute_generator as data

html(t"""<button {data.on("click", "console.log('clicked')").debounce(1000).stop}>My Button</button>""")
```

ianjosephwilson:
I agree that it is a little too strict.  `Mapping` does seem more Pythonic.  Our intention was that because we are performing runtime checks we didn't want to make them too slow or too hard to reason about.  We use `Sequence` in another situation, for a `class` attribute, so it seems that `Mapping` is a more consistent parallel.  

I'll look into this and see if we can adjust this.  We should probably do it in other places as well to maintain consistency.  

Thanks for this clear use-case and I'll see if we can adjust things.

davepeck:
@gazpachoking @ianjosephwilson I just merged PR #147 and pushed update to PyPi (v0.1.18). Thanks so much!

gazpachoking:
Thanks guys!

PR #147, opened 2026-07-08
Use `Mapping` in place of `dict` for `isinstance` check in "spread", data, aria, class and style attributes.

ianjosephwilson:
Most of the builtins are subclasses of `dict` but libraries might implement their own dict-like implementations.  This seems more Pythonic, and we already use `Sequence`, `isinstance(v, Mapping)` is slower but we'll have to tackle that some other way.  Possibly with some sort of attribute caching scheme that would bypass all the checks anyways.

We use `UserDict` (which is not a subclass of `dict`) to test this in a few tests but adding it to every case feels a little redundant since we are using a regular dict in the tests which is a `Mapping` already.

We already should not be mutating the user's input anyways so `Mapping` is a good for that check as well.

### DIFF
--- a/tdom/processor.py
+++ b/tdom/processor.py
@@ -1,5 +1,5 @@
 import typing as t
-from collections.abc import Callable, Iterable, Sequence, Mapping
+from collections.abc import Callable, Iterable, Mapping, Sequence
 from dataclasses import dataclass, field
 from functools import lru_cache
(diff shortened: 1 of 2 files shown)


ianjosephwilson:
Potential fix for #146

davepeck:
@ianjosephwilson LGTM!

PR #150, opened 2026-08-04
Consolidate placeholder tracking in the source tracker.

ianjosephwilson:
- Morph `SourceTracker` into an Iterator.
- Move placeholder injection (and tracking) internally into the `SourceTracker`.
- Redirect `TemplateParser`'s placeholder extraction to proxy methods of the `SourceTracker.  The parser still needs to know when its just looking at a snapshot versus when it is "marking" the placeholder as found.  Not ideal but this composition feels better and more distinct from the regular parsing duties.
- Use `configure_source_tracker` factory function because later we bind the `PlaceholderConfig` to members other than `PlaceholderState` that depend on the size of the placeholders (defined by the config).

The `SourceTracker` methods `get_expression` and `format_starttag` will be removed in another PR and those responsibilities will be in the `SourceReader` (constructed and accessed during exceptions).

### DIFF
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -1,10 +1,16 @@
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 from dataclasses import dataclass, field
 from html.parser import HTMLParser
-from string.templatelib import Interpolation, Template
+from string.templatelib import Template
 
 from .htmlspec import VOID_ELEMENTS
-from .placeholders import PlaceholderState
+from .placeholders import (
+    PlaceholderConfig,
+    PlaceholderState,
+)
+from .placeholders import (
+    make_placeholder_config as default_make_placeholder_config,
+)
 from .template_utils import TemplateRef, combine_template_refs
 from .tnodes import (
     TAttribute,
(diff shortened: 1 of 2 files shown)


ianjosephwilson:
@davepeck Thanks for merging those changes.  Here is the next "chunk" that continues to move things into alignment with the error messaging.

davepeck (commented):
LGTM! Two _very_ small and ignorable comments:

1. There are a couple places in `parser.py` where we do `if self.source:` or `if self.source and ...:`. Maybe these should become `if self.source is not None` for future safety (it's fine now, but what happens if we add, say, `__len__`?)

2. Should we have a variant of `test_iter` that tests `t""` and `t"one-string-only"` cases? 

As for `find_placeholders` and `remove_placeholders`: I think it works, but agree you have to be cautious.

Anyway, extremely minor feedback! LMK when you're good for a merge.

davepeck (approved):

ianjosephwilson:
@davepeck I added a parametrized test for those simple string cases you mentioned.  If this looks ok I think you should merge it so I can start prep'ing the next part.

davepeck:
@ianjosephwilson Merged; thanks!

By the way, one thing I've been contemplating is moving the source structure to something more standard in the python world, specifically:

`src/tdom/...` <-- top level `src/` which seems to be the thing to do these days
`tests/...` <-- top level `tests/` that mirror the structure of `src/tdom/` aka `src/tdom/sub/whippersnapper.py` gets `tests/sub/test_whippersnapper.py`

This is just a bunch of churn, so I want to avoid it until you are in a position to take it. If that's now, cool; if that's, like, after the error handling stuff is settled, cool too -- just let me know. (And if you don't like this structure, LMK what you'd prefer!)

ianjosephwilson:
> @ianjosephwilson Merged; thanks!
> 
> By the way, one thing I've been contemplating is moving the source structure to something more standard in the python world, specifically:
> 
> `src/tdom/...` <-- top level `src/` which seems to be the thing to do these days `tests/...` <-- top level `tests/` that mirror the structure of `src/tdom/` aka `src/tdom/sub/whippersnapper.py` gets `tests/sub/test_whippersnapper.py`
> 
> This is just a bunch of churn, so I want to avoid it until you are in a position to take it. If that's now, cool; if that's, like, after the error handling stuff is settled, cool too -- just let me know. (And if you don't like this structure, LMK what you'd prefer!)

Haha don't you dare DAVE!  Thanks for asking and I think these are both good ideas for once we get the error stuff in assuming nothing else is going on.  But it would probably be safer to wait in case I need to scissor-and-glue-stick together more sub-sub PRs for us to digest to get this thing in.

davepeck:
> But it would probably be safer to wait

Yup, no rush on this one. Will do later.

PR #155, opened 2026-08-09
Compute and store position of TNodes during parsing for improved error messages.

ianjosephwilson:
## Design Notes

Goals
- don't overload the cache (and don't make the current cache incompatible)
- don't add a bunch of info to the core apis that we have to work around
- allow the tracking to be disabled (potentially)

#### Translation

`(template_part_index, offset_in_template_part) = translate(parser_line_number, offset_in_parser_line)`

The underlying `HTMLParser` provides a line position via `getpos()` that is relative to the entire template flattened into a continuous string with interpolations replaced with placeholders.  That line position is translated to a "part" position in the `Template` using unified indexing (combined strings and interpolations).  This part position applies to all templates with the same tuple of strings.

Note that the purpose of the "unified indexing" (all this `2*len(t.strings) - 1` or `index // 2` madness) is that it allows for exact storage of the part location and also has less complicated usage and more direct iteration.  This is especially true when starting at a part and ending at a part compared to storing and trying to iterate from a `s_index` or an `i_index`.

#### Source Position and Info

The translation from line position to part position is performed for all nodes during parsing and then stored with that tnode.  For tags (elements and components) more info is wanted.  This extra source info is stored in a lookup to avoid overloading the cache and core apis.  A new `TTree` structure is used to store the `root: TNode` and the lookup packed as `sinfos`.  This keeps the core `TNode` subclasses simple as they only have one extra field: `source_pos: PartPosition`.  This assumes that if a node needs extra source info then its source position is unique within that template which seems sensible for now.

## Next Steps

Assuming we don't hit any brick wall blockers we'd keep going along until we get to semi-functioning exception handling:

- part 4: add `TemplateRef` slicing
- part 5: add `SourceReader` and printable repr of `Template`s
- part 6: parser exceptions and handling
- part 7: processor exceptions and handling

### DIFF
--- a/tdom/parser.py
+++ b/tdom/parser.py
@@ -4,6 +4,11 @@ from html.parser import HTMLParser
 from string.templatelib import Template
 
 from .htmlspec import VOID_ELEMENTS
+from .parser_utils import (
+    HTMLAttribute,
+    ParserPositionTranslator,
+    make_parser_pos_translator,
+)
 from .placeholders import (
     PlaceholderConfig,
     PlaceholderState,
@@ -11,8 +16,10 @@ from .placeholders import (
 from .placeholders import (
     make_placeholder_config as default_make_placeholder_config,
 )
-from .template_utils import TemplateRef, combine_template_refs
+from .source import LinePosition
+from .template_utils import PartPosition, TemplateRef, combine_template_refs
 from .tnodes import (
+    TagSourceInfo,
     TAttribute,
     TComment,
     TComponent,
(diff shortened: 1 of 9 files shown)


ianjosephwilson:
@davepeck Thanks for merging the last part.  This part is ready to review.

davepeck:
@ianjosephwilson Thanks for your patience on this one. 

Overall, it's great. Two pieces of feedback, one minor, one medium:

1. We should mark `TTree` as `@dataclass(frozen=True, slots=True)`.
2. I spent a bunch of time pondering `ParserPosition` and `ParserPositionTranslator` and have an alternative implementation suggestion: instead of keeping track of `line`/`offset`/`eol`/`eof`, things seem to fall out simpler if we just use the absolute parser position (a single `int`!). I also noticed that there's effectively an `O(n^2)` behavior in the current way the `TemplateParser` and `ParserPositionTranslator` interact; my alternative implementation follows a suggestion you had in a comment, and precomputes line start and part end offsets. I think the overall behavior becomes `O(n log n)` after that. The public interface (`ParserPositionTranslator.translate(...)`) doesn't change; this is all implementation detail. See [my suggested implementation of `parser_utils.py` here](https://github.com/t-strings/tdom/blob/a933049d8d0287249ac8d02fbf752cb78b3f6460/tdom/parser_utils.py). (`MutableLinePoisition` can go away after that, too, I think?)

ianjosephwilson:
@davepeck For maximum confusion I tried to pull in your commits and then add some name adjusting on top of it.  There is no right answer here though because there are so many similar concepts in the same module: offsets, lengths, positions, absolutes, parts, lines.  I can't tell if re-arranging the names just "feels" better but reading it anew it would be just as confusing.  Oh well, this feels a little bit more consistent to me.   

I think the absolute position/offset concept is great and should just fit right in as a replacement like you said and not affect the rest of the changes (at a high level).

Let me know what you think or just merge this in.

davepeck:
@ianjosephwilson haha, nice. It looks good to me and I like your changes. I'm mergin'!

ianjosephwilson:
Awesome, thanks!  That was a big one.

Issue #156, opened 2026-08-17
Support passing through arbitrary data- attributes through component functions


gazpachoking:
It would be nice if there were some way to pass through arbitrary arguments with component functions. It works fine now for normal attributes, but aria- and data- attributes will need to have underscores replaced with dashes to pass through, which is annoying to remember on every component function. Even worse, both dashes and underscores are valid characters in a data- attribute, so it's impossible to restore them even with special handling right now. (I'm thinking specifically of Datastar attributes at the moment, but it's a problem in general) e.g.
```python
import tdom

def Tag(**args):
    return t"<div {args}></div>"

print(tdom.html(t"""<{Tag} data-signals:my-signal__ifmissing="1" />"""))
```
This would produce `<div data_signals:my_signal__ifmissing="1"></div>` with no way for the component to know which underscores should have been replaced with dashes.

Is there some way I'm not thinking of to handle this already? I don't have the full solution, but ideally `**args` would collect all the remaining attributes without the dash to underscore replacements for easy pass-through.

ianjosephwilson:
I think I understand what you mean.  Thanks for showing a real example.  I think we *could* make this work but it is a little confusing to have the attributes change depending on if named, ie. `def Tag(any_attr)` versus in `**kwargs`.  So I'm not sure if we *should*.  I'll have to think about it and look back at what we were thinking.  The common case of just any attribute is almost more weird, ie. `any-attr` becomes `any_attr` in `**kwargs`.

I do know that one thing we were thinking was that if you need specialized attributes then the component should take over the responsibility of handling them itself.  I tried to show that in `test_specialize_attrs` below.

```python
from tdom import html
from string.templatelib import Template


def test_any_attr() -> None:
    def Tag1(**args: dict[str, str]) -> Template:
        return t"<div {args}></div>"
    res = html(t"""<{Tag1} any-attr="1" />""")
    # This is weird, ...deterministic, but still weird.
    assert res == '<div any_attr="1"></div>', f'{res=}'


def test_specialize_attrs() -> None:
    def Tag2(signals: dict[str, str]|None = None) -> Template:
        if signals is None:
            signals = {}
        # The component pre-serializes these specialized attributes itself.
        signal_attrs = {f'data-signals:{k}': v for k, v in signals.items()}
        return t"<div {signal_attrs}></div>"
    signals = {"my-signal__ifmissing": "1"}
    res = html(t"""<{Tag2} signals={signals} />""")
[...]

PR #157, opened 2026-08-18
Allow extra component params to pass through via `kwargs` as-is.

ianjosephwilson:
Potential fix for #156 

Instead of trying to handle the various different situations where a collision occurs an exception is raised.  The snake-ifying of the params is already an ergonomic feature and I think cases where the names collide would likely be a mistake and an exception aligns with those same ergonomics.

### DIFF
--- a/tdom/processor.py
+++ b/tdom/processor.py
@@ -432,10 +432,7 @@ def _prep_component_kwargs(
     # Add all supported attributes
     for attr_name, attr_value in attrs.items():
         snake_name = _kebab_to_snake(attr_name)
-        if snake_name in callable_info.named_params and (
-            # Don't use a coerced snake name if there is another snake attr available.
-            snake_name == attr_name or snake_name not in attrs
-        ):
+        if snake_name in callable_info.named_params:
             if snake_name in kwargs:
                 raise ValueError(
                     f"Ambiguous attribute {attr_name}: Two attributes resolved to the same named param {snake_name}."
--- a/tdom/processor.py
+++ b/tdom/processor.py
@@ -432,10 +432,19 @@ def _prep_component_kwargs(
     # Add all supported attributes
     for attr_name, attr_value in attrs.items():
         snake_name = _kebab_to_snake(attr_name)
-        if snake_name in callable_info.named_params or callable_info.kwargs:
+        if snake_name in callable_info.named_params and (
+            # Don't use a coerced snake name if there is another snake attr available.
+            snake_name == attr_name or snake_name not in attrs
+        ):
+            if snake_name in kwargs:
+                raise ValueError(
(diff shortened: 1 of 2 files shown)


gazpachoking:
It's a tiny bit weird that duplicating attributes will result in an exception if they are named parameters, but not otherwise. The other possibility I can think of is to silently throw away the later repeated attributes similar to what the browser would do. I'd probably switch to `kwargs.setdefault(attr_name, attr_value)` in the non-named parameter case rather than overwriting for the same reason of sticking closer to what the browser would do with invalid repeated attributes. (I'm pretty sure the browser keeps the first one rather than the last.)

davepeck:
I think this is a good tweak to our weird HTML-attributes-meet-python-functions boundary. 

I'm tempted to merge it as-is. It may not be the final version, but at least it addresses #156 today.

I _think_ I'm okay with this?: `t"<{Component} foo-bar=42 foo_bar=99 />"` will work if we `def Component(**kwargs): ...` (`kwargs` will contain both keys) but will raise if we `def Component(foo_bar: str, **kwargs): ...` instead.

-----

We might want to separately revisit tdom's handling of duplicate attributes. tdom has a last-wins behavior, which is useful for composition:

```python
def Component(**kwargs):
    return t"<div {kwargs}></div>"
    
defaults = {"title": "Default", ...}
html(t'<{Component} {defaults} title="Specific" />')
# '<div title="Specific"></div>'
```

I _think_ this matches what most users would expect. (It's not clear to me that tdom should even try to act like a browser in most cases.)

The current implementation may have some strange edges, though:

```python
html(t'<div title="First" title="Second"></div>')
'<div title="Second"></div>'
```

ianjosephwilson:
> We might want to separately revisit tdom's handling of duplicate attributes. tdom has a last-wins behavior, which is useful for composition:
> 
> ```python
> def Component(**kwargs):
>     return t"<div {kwargs}></div>"
>     
> defaults = {"title": "Default", ...}
> html(t'<{Component} {defaults} title="Specific" />')
> # '<div title="Specific"></div>'
> ```

This seems to further justify the exception approach because the entire attribute resolution pipeline runs BEFORE the kebab mangling/coercion.  We almost have too much going on already that probably isn't obvious.  I think this PR is a good step and later on maybe we would hopefully have a better options.

davepeck:
> This seems to further justify the exception approach because the entire attribute resolution pipeline runs BEFORE the kebab mangling/coercion. We almost have too much going on already that probably isn't obvious. I think this PR is a good step and later on maybe we would hopefully have a better options.

Agree on all fronts. Okay if I merge?

ianjosephwilson:
Yes, please merge.  Thanks.

PR #158, opened 2026-08-19
Explore a different parser translate approach

davepeck:
Suggested implementation of `parser_utils.py` on top of #155 that:

1. Only uses absolute parser positions, removing the need for a bunch of edge-case math.
2. Precomputes line start offsets, which takes us from effective n^2 behavior to n log n

### DIFF
--- a/tdom/parser_utils.py
+++ b/tdom/parser_utils.py
@@ -45,7 +45,7 @@ def make_parser_pos_translator(
     )
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class ParserPositionTranslator:
     line_start_offsets: tuple[int, ...]
     """Absolute offsets where lines in the parser input start."""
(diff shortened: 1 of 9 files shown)


ianjosephwilson:
This approach looks great, maybe even "neato".  The `bisect_left` is very slick.  I wasn't as concerned about the `n^2` but we needed some major bug spray with my prior approach.  I think you were on your way but it seems that this just needs a little bit of polish.  I had a few clarifying questions / minor things I'm going to put in comments but here is some naming polish I was thinking:

Instead of using `type ParserPosition = int` you could just treat the new "absolute" concept as its own type for now: `type AbsolutePosition = int`.  Mainly just for our internal use in this module because this code is so dense.

Then maybe realign the signatures to something like this:

```python
def line_pos_to_abs_pos(
        self,
        line_pos: LinePosition,
    ) -> AbsolutePosition: ...
def abs_pos_to_part_pos(self, abs_pos: AbsolutePosition) -> PartPosition: ...
def translate(self, parser_pos: LinePosition) -> PartPosition: ...
```

tdom/parser_utils.py
```diff
@@ -83,70 +37,54 @@ def make_parser_pos_translator(
...
+    line_start_offsets: tuple[int, ...]
+    """Absolute offsets where lines in the parser input start."""
 
-    source_text_lines: tuple[str, ...]
-    " The source text of the entire template, with placeholders. "
-
-    line_to_part_pos: dict[int, PartPosition]
-    " Precomputed mapping from line number to part position. "
+    part_end_offsets: tuple[int, ...]
+    """Cumulative end offsets of the placeholder-expanded template parts."""
```
ianjosephwilson, line 54: These are also absolute offsets/positions right?

davepeck: Yes. Any suggestions for how say that in the comment are most welcome. :-)

davepeck:
(Closing now that @ianjosephwilson has merged in to #155)

PR #160, opened 2026-08-22
Add template ref slicing.

ianjosephwilson:
Allow slicing of template refs from start part to end part.    

Interpolations are indivisible but string "parts" can be further sliced with an `offset`/`limit`.

The motivation for this is to cut out part of a `Template` to put into an error message.


tdom/template_utils.py
```diff
@@ -92,6 +92,81 @@ def resolve(self, interpolations: tuple[Interpolation, ...]) -> Template:
...
+                strings.append("")
+            return TemplateRef(strings=tuple(strings), i_indexes=tuple(i_indexes))
+        else:
+            if first % 2 == 0:
+                strings.append(self.strings[first // 2][offset:])
+            else:
+                # offset == 0, so template-ify with empty string but start by
+                # including this interpolation.
+                strings.append("")
+                i_indexes.append((first - 1) // 2)
```
davepeck: Should this be `i_indexes.append(self.i_indexes[(first - 1) // 2])`?

tdom/template_utils.py
```diff
@@ -92,6 +92,81 @@ def resolve(self, interpolations: tuple[Interpolation, ...]) -> Template:
...
+            if index % 2 == 0:
+                if index == last:
+                    strings.append(self.strings[index // 2][:limit])
+                else:
+                    strings.append(self.strings[index // 2])
+            else:
+                if index == last:
+                    break  # offset == 0, so exclude this interpolation.
+                else:
+                    i_indexes.append((index - 1) // 2)
```
davepeck: And likewise here, `i_indexes.append(self.i_indexes[(index - 1) // 2])`?

davepeck (commented):
Thanks, Ian! 

Excited to get this next piece in. I'll do a full review soon, but for the moment: I think the tests may (by coincidence) be obscuring a bug here?

Which makes me wonder... is there ever a real case where a `TemplateRef.i_indexes` isn't contiguous? We have test cases, of course, but I wonder if a better representation would be `TemplateRef.i_start`, and then have `TemplateRef.i_stop` be computed as `self.i_start + len(self.strings) - 1`. I wonder if a few things would end up cleaner as a result. At the very least we wouldn't have confusing weirdness like I noticed here. Hmmm... need to think about it.

tdom/template_utils.py
ianjosephwilson: Great catch.  That's a bad one too.  Or would be hard to debug later on.

ianjosephwilson:
> Which makes me wonder... is there ever a real case where a `TemplateRef.i_indexes` isn't contiguous? We have test cases, of course, but I wonder if a better representation would be `TemplateRef.i_start`, and then have `TemplateRef.i_stop` be computed as `self.i_start + len(self.strings) - 1`. I wonder if a few things would end up cleaner as a result. At the very least we wouldn't have confusing weirdness like I noticed here. Hmmm... need to think about it.

Yeah I think I need to re-review it myself tomorrow.  I think this strategy is still good but there is some unwritten meta thing going on here that is hard for me to even explain..?  We are thinking of `PartPosition` as an index into `template.interpolations` (and `template.strings`) but then we build out `TemplateRef.i_indexes` and now we are using `PartPosition` also to index `tref.i_indexes` which is a subtle difference.  I think the real crux is that if you called `slice_to_tref` ONLY you'd be fine because the `i_indexes` are ALWAYS 1-to-1 as constructed but if you called `TemplateRef.slice` directly then the `i_indexes` are not necessarily 1-to-1 depending on where that `TemplateRef` came from.  That's the real confusion.  I think we want the behavior, but it needs to be spelled out somehow.

davepeck:
> but there is some unwritten meta thing going on here

Yeah, I think so. I don't totally have my head wrapped around it yet, but one instinct is that we're using `TemplateRef` in at least some places where it feels like two `PartPosition`s might be the simpler representation. (Call it a `PartRange`? `TemplateSpan`?) And, if we had that abstraction, I suppose `slice()` would belong on _it_? (And would it slice `Template` only, maybe?)

Dunno, still fuzzy, so going to poke at it a bit.

davepeck:
Update: poked at it and a whole bunch of stuff unravelled; maybe too much? Dunno, but the exploration is at #161

davepeck:
Closing now that #161, which built on this directly, is merged.

PR #161, opened 2026-08-22
Implement `TemplateSpan`, `extract()`, and clarify relation to `TemplateRef`

davepeck:
Building on Ian's work in PR #160.

1. Moved several sanity checks at the beginning of `slice()` to a new `PartPosition.__post_init__`, which feels more natural to me.
2. Made `PartPosition` a `@dataclass(order=True)`; this makes the _other_ sanity checks in `slice()` easier to read, I think.
3. Removed the loop in `slice()`. I realized that it's only the boundaries that matter here; everything else passes through. Hopefully this is simpler!
4. Replaced `TemplateRef.i_indexes` with `TemplateRef.i_start` and a _computed_ `i_stop`; enforce that `TemplateRef` is contiguous everywhere. This actually had the nice side effect of allowing us to remove `to_naive_template()`, `from_naive_template()`, and move what was left of `combine_template_refs` to `TemplateRef.concat()`.
5. The big one! I think I may have successfully teased apart `TemplateRef` and `PartPosition`? See comments below.


davepeck:
On #160 @ianjosephwilson  said:

> but there is some unwritten meta thing going on here

Yes! Ian, this gave me weird dreams. :-)

It felt like `PartPosition` and `TemplateRef` were colliding. There were places where we used `TemplateRef` even though a pair of `PartPosition`s seemed like the cleaner representation. For a moment, I wondered whether we could get rid of `TemplateRef` entirely.

I started by introducing `TemplateSpan`, which holds `start` and `stop` positions. I moved `TemplateRef.slice()` there and renamed it `extract()`, mostly because `span.slice(t)` felt backward to me. But I kept getting confused: should `extract()` take a `Template` or `TemplateRef` as input? What should it return? Did we want both variants?

After futzing with it, I finally understood when to use `TemplateRef` and when to use `TemplateSpan`. `TemplateRef` holds on to its own strings, including strings produced by the parser, like `<` instead of `&lt;`, and lets us defer binding the real interpolations. That makes it safe to store behind the cache and bind against the current template later. On the other hand, `TemplateSpan` doesn't capture content. It just remembers a range and, importantly, that range should _always_ be relative to the original template, never to parser-produced strings.

This meant that `extract()` should take a `Template` and return a new `Template`; `TemplateRef` shouldn't be involved. It also meant that we should use `TemplateSpan` for component children and start-tag source information, since those need to be locations in the original source. This fixed some subtle bugs where we were interpreting positions against parser-produced strings.

(One wrinkle: `get_starttag_ref()` had to become `get_starttag_span()`, which meant we needed to compute the endpoint. There's a new `LinePosition.advance_over()` method to help.)

Curious your take on this, @ianjosephwilson!

davepeck:
If we like this direction, proposed future renaming: `PartPosition` -> `TemplatePosition`.

_Maybe_ there's a better name for `TemplateRef`, too? `UnboundTemplate`? `TemplateBones`? `SpookyTemplateSkeleton`? I dunno.

tdom/parser.py
```diff
@@ -79,10 +81,8 @@ class OpenTFragment:
 @dataclass
 class OpenTComponent:
     start_i_index: int
-    children_start_s_index: int
-    """The strings index where the component's children template starts."""
-    offset_into_children_start_s: int
-    """The offset INTO the starting string where the component's children template starts."""
+    children_start: PartPosition
+    """Source position where the component's children start."""
```
ianjosephwilson, line 85: This is a nice pickup.

tdom/parser.py
```diff
@@ -305,38 +304,21 @@ def make_open_tag(
...
+        # This must be called while handling the tag because HTMLParser retains
+        # only the most recently parsed start tag text.
+        starttag_span = self.get_starttag_span()
+        source_pos = starttag_span.start
 
         return OpenTComponent(
             start_i_index=i_index,
-            children_start_s_index=children_start_s_index,
-            offset_into_children_start_s=offset_into_children_start_s,
+            children_start=starttag_span.stop,
```
ianjosephwilson, line 326: Another great pickup here too.

tdom/parser.py
```diff
@@ -370,79 +351,25 @@ def finalize_tag(
...
                 sinfo=sinfo,
             ):
-                children_ref = self.extract_component_children_ref(
-                    start_i_index=start_i_index,
-                    endtag_i_index=endtag_i_index,
-                    children_start_s_index=children_start_s_index,
-                    offset_into_children_start_s=offset_into_children_start_s,
-                    template=source.template,
+                children_span = (
+                    TemplateSpan(start=children_start, stop=endtag_pos)
```
ianjosephwilson, line 370: Amazing again.

tdom/parser.py
```diff
@@ -475,18 +402,20 @@ def validate_end_tag(self, tag: str, open_tag: OpenTag) -> int | None:
...
-        source = self.get_source()
+    def get_starttag_span(self) -> TemplateSpan:
+        """Return the source span occupied by the current start tag."""
         starttag_text = self.get_starttag_text()
         if starttag_text is None:
             raise AssertionError("Expected the parser to have starttag_text set.")
-        return source.find_placeholders(starttag_text)
+
+        parser_start = self.get_parser_pos()
+        parser_stop = parser_start.advance_over(starttag_text)
```
ianjosephwilson: I'm not sure why but this feels little awkward.  Maybe its the only place where we move the parser position ourselves rather than relying on `getpos`?  I guess if the parser `getpos()` is correct and their `starttag_text` is correct then `pos = getpos() + starttag_text` *should* be correct.

tdom/processor.py
```diff
@@ -337,7 +335,7 @@ def _resolve_t_attrs(
                 else:
                     new_attrs[name] = attr_value
             case TTemplatedAttribute(name=name, value_ref=ref):
-                attr_t = ref.resolve(interpolations)
+                attr_t = ref.bind(template)
```
ianjosephwilson: I'm not sure if it  really matters but we could probably keep sending only the interpolations here because I don't think we use the `.strings`.

ianjosephwilson (commented):
Overall this seems better.  I'm surprised we don't depend on runs of interpolations that are not continuous.  It makes sense to specialize in the context of the slicing because that was an issue, ie. "How generalized should this be?"  I noted a few things that seem to be "coming together" nicely.  I'm not sure how python caching works w.r.t. interning but its possible that there is some savings by storing the spans instead of the tuples of `strings` and `i_indexes` in certain cases as well.  Although we might watch out for situations where pre-cutting the strings might be important.

I agree that we might play with the names for a bit longer but nothing that is going to block things.  We might find more clarity as the error messages generation / source extraction starts coming into play.

ianjosephwilson:
> If we like this direction, proposed future renaming: `PartPosition` -> `TemplatePosition`.
> 
I vote to leave this for now because once we use the source text more we almost have a 3rd concept of positioning within the template source (ie. expressions, etc.) and we should re-consider it all together.

> _Maybe_ there's a better name for `TemplateRef`, too? `UnboundTemplate`? `TemplateBones`? `SpookyTemplateSkeleton`? I dunno.

Not to throw wrenches into the gears of progress ... but I wonder if we might actually promote TemplateRef to TemplateSpan and TemplateSpan itself becomes TemplateSlice?  This mirrors the concept of python's `Slice` object:
https://docs.python.org/3/library/functions.html#slice

tdom/processor.py
davepeck: Agreed, I like the previous interpolations-only shape better. Fixed.

tdom/parser.py
davepeck: > feels a little awkward

Yeah, agree. I sorta slammed it in at the last minute. Good call on `starttag_text`; see how you like my update in `7fc9000` instead.

davepeck:
> I vote to leave this for now

Agreed. I think, stepping back, I'm just going to leave the names as they've settled down. We can revisit later if we don't like 'em.

--- 

Okay, I made some updates. How do we feel about these changes now?

tdom/parser_utils.py
```diff
@@ -129,3 +129,22 @@ def translate(self, parser_pos: LinePosition) -> PartPosition:
         """
         abs_pos = self.line_pos_to_abs_pos(parser_pos)
         return self.abs_pos_to_part_pos(abs_pos)
+
+    def translate_span(
+        self,
+        parser_start: LinePosition,
+        parser_length: int,
```
ianjosephwilson, line 136: I think the concept of absolute locations within the placeholder embedded source is still being digested by my brain.  I didn't understand what you were talking about regarding the "raw" text earlier that wasn't changed by the parser but I see it now with this fix.  I initially thought it wouldn't work but I get it now.  The `starttag_text` should be unaltered/raw and so we can use that to adjust the absolute location because all the absolute locations are based on the initial unaltered/raw text.  The world of text manipulation is like a footgun armory.  

I like this fix better but I think we still need to hone the names a bit.  Maybe this is something we could handle later when things come together.  

We sort of have a conflict of naming between the "units", (line+offet, char, parts) and the material being measured (placeholder-expanded input, post parsed text, the template source w/ expressions, etc.).  This is related to the Template* classes as well.  The containers holds different forms but also the contents can be of different material.  

I think you should ship this as-is though I'm just keeping the brain storming going, maybe "name storming"?

ianjosephwilson:
@davepeck Thanks for grinding through all this.  Brain-load my comment and merge it and I can rebase the next chunk of my mega PR on this.

tdom/parser_utils.py
davepeck: > The world of text manipulation is like a footgun armory.

hahaha

> We sort of have a conflict of naming between the "units", (line+offet, char, parts) and the material being measured (placeholder-expanded input, post parsed text, the template source w/ expressions, etc.). 

Agree; I don't think it's fully teased apart yet &mdash; in the code _or_ in my head. But hopefully at least the code is trending in the right direction. (I refrain from comments on my head.) I don't think I _ever_ noticed the raw vs entity-expanded text subtlety until you got us poking this particular bear.

davepeck:
@ianjosephwilson brain-loaded -- and merging. Thanks!

PR #162, opened 2026-08-24
Simplify `PartPosition`

davepeck:
Instead of `PartPosition.index` representing an index into interleaved strings and interpolations, we now have `PartPosition.s_index` which can only _ever_ represent an index to a Template string. An `offset` of `0` means "just before this string" (and therefore "just after the preceding interpolation, if any"); an `offset` of `len(the_string)` means "just after this string" (and therefore "just before the following interpolation, if any").

This simplifies a bunch of downstream code (see `extract()` and `abs_pos_to_part_pos()` for example) and removes some of the awkwardness around `offset` if the position were previously an interpolation.

### DIFF
--- a/tdom/parser_utils_test.py
+++ b/tdom/parser_utils_test.py
@@ -36,9 +36,9 @@ class TestParserPositionTranslator:
         placeholder_length = len(ph_config.make_placeholder(0))
 
         assert ppt.abs_pos_to_part_pos(0) == PartPosition(0, 0)
-        assert ppt.abs_pos_to_part_pos(1) == PartPosition(1, 0)
-        assert ppt.abs_pos_to_part_pos(1 + placeholder_length) == PartPosition(2, 0)
-        assert ppt.abs_pos_to_part_pos(2 + placeholder_length) == PartPosition(2, 1)
+        assert ppt.abs_pos_to_part_pos(1) == PartPosition(0, 1)
+        assert ppt.abs_pos_to_part_pos(1 + placeholder_length) == PartPosition(1, 0)
+        assert ppt.abs_pos_to_part_pos(2 + placeholder_length) == PartPosition(1, 1)
 
         with pytest.raises(ValueError, match="Absolute position falls outside"):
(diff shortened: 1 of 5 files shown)


davepeck:
@ianjosephwilson for your consideration

tdom/parser_utils_test.py
```diff
@@ -63,75 +63,73 @@ def test_translate_span(self, ph_config):
     def test_case_nontailing_string_ends_with_newline(self, ph_config):
         ppt = make_ppt(t"a\n{0}b", ph_config)
         assert ppt.translate(LinePosition(line=2, offset=0)) == PartPosition(
-            index=1, offset=0
-        ), """This could also be considered PartPosition(index=1, offset=0)
-        but either way should work. """
+            s_index=0, offset=len("a\n")
+        ), "The placeholder starts at the end of its preceding string."
```
ianjosephwilson, line 67: This does seem to solve this sort of ambiguity which is great.

tdom/template_utils_test.py
```diff
@@ -278,8 +270,8 @@ def test_interval(
 
     def test_interpolation_interval(self) -> None:
         extracted = TemplateSpan(
-            start=PartPosition(1),
-            stop=PartPosition(2),
+            start=PartPosition(0, len("<div>")),
```
ianjosephwilson, line 273: This I think is the weirdness part but the concepts (and math) still all seem to fit together.

ianjosephwilson:
@davepeck  This does seem more elegant overall than what I'd call my brute force approach.  There is one weird situation but the math works and the ambiguous case seems resolved.  I'm not tied to it so you should just merge it.  We can keep adjusting things.

The one thing I do prefer is iterating over entire template with the 2-based index whenever that is needed but it seems like more and more of those loops are being replaced with more specialized solutions.  Maybe we'll see what is in there when the smoke clears from all this.

tdom/parser_utils_test.py
davepeck: Yeah, this had been irking me.

PR #163, opened 2026-08-24
Add source reader and template repr.

ianjosephwilson:
- Adds `SourceReader` class.
    - Created when an exception occurs to read back the source for error messages.
- Adds `MutableLinePosition` class.
    - Accumulates the line position (w/ newlines) through interpolation expressions before showing the line/offset in error messages.
- Adds `interpolation_repr()`, `template_repr_iter()` and `template_repr()` functions.
    - Utility functions for helping `SourceReader`.

This code is sort of disconnected until we add in the actual exception/error handling.

I was able to sub out the slicing concept for the `TemplateSpan().extract()` pretty easily but other parts might change `ref` for `span` but I'm not sure until we add the next chunk.

### DIFF
--- a/tdom/source.py
+++ b/tdom/source.py
@@ -1,4 +1,8 @@
+import typing as t
 from dataclasses import dataclass
+from string.templatelib import Interpolation, Template
+
+from .template_utils import PartPosition, TemplateRef, TemplateSpan
 
 
 @dataclass(slots=True, frozen=True)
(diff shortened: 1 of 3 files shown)


ianjosephwilson:
@davepeck *Almost* to the parser exceptions.  It is amazing how much code  and "shoehorning" all these details take.

tdom/source.py
```diff
@@ -9,3 +13,103 @@ class LinePosition:
...
+
+    def make_interpolation_repr(self, i_index: int) -> str:
+        return interpolation_repr(self.template.interpolations[i_index])
+
+    def to_template_pos(self, source_pos: PartPosition) -> LinePosition:
+        """
+        Convert a (template) part position into a line position based on the
+        string representation of the template.
+        """
+        pos = MutableLinePosition()
```
davepeck: I love your `template_repr()` and it feels like it can do the hard work for us:

```python
def to_template_pos(self, source_pos: PartPosition) -> LinePosition:
    span_up_to_pos = TemplateSpan(start=PartPosition(0, 0), stop=source_pos)
    repr_up_to_pos = template_repr(span_up_to_pos.extract(self.template))
    lines_up_to_pos = repr_up_to_pos.split("\n")
    return LinePosition(line=len(lines_up_to_pos), offset=len(lines_up_to_pos[-1]))
```

That would also allow us to delete `MutableLinePosition` which I'm realizing we maybe don't even need, now _or_ in the future?...

tdom/source.py
```diff
@@ -9,3 +13,103 @@ class LinePosition:
...
+    return f"{{{expr_str}{conversion_str}{format_spec_str}}}"
+
+
+@dataclass
+class SourceReader:
+    """Format report-like strings from template source for error reporting."""
+
+    template: Template
+
+    def values_match(self, i_index1: int, i_index2: int) -> bool:
```
davepeck, line 55: This feels strange to me on `SourceReader` although I don't have a better suggestion at the moment. Is this ultimately about components? I wonder if we'll want a variant that is just for them, more localized to that code?

Also, `==` is important but I suppose it _could_ raise an exception since it could be arbitrary user code. I wonder if (a) we could do a faster/safe `is` check first, and (b) we should catch exceptions here. (?)

davepeck (commented):
Awesome, thanks @ianjosephwilson!

Some comments here to consider but overall a nice narrow slice.

PS: I merged in #162 and updated the tests in this PR to use the new coordinates.

tdom/source.py
ianjosephwilson: I agree, there are some methods that come in later on the parser that feel out of place as well.  We might want another helper class/module that contain all these, like SourceInspector or something.  I guess I don't usually make this many classes but the idea would be to make accessing the `template` more explicit and put together these oddities.  Also yes it has to do with components and everybody's favorite ambiguous trailing slash.  So when there is an exception we go digging around to see why and we have to "read the source template" (or inspect?) to check these values.

tdom/source.py
ianjosephwilson: "First he came for the naive templates and then he came for the `MutableLinePosition`!"  It is confusing to have both classes for sure, at one point they were the opposite, LinePosition and FrozenLinePosition or something.  I think this is the last place it was used...

davepeck:
@ianjosephwilson Looks good to me. LMK if you're ready for a merge. Thanks!

ianjosephwilson:
@davepeck Should be ready to merge.  I'll go back over the coverage at the end because some of these small methods we might just remove.

PR #165, opened 2026-08-26
Possible Refactor For Parser Error Messages

ianjosephwilson:
@davepeck This is a possible refactor of the other branch ( #164 ) where we pull out all the non-essential error handling into a helper class.  Sorry I had sent the other PR yesterday from what I've been rebasing/rehashing but I got restless and started fiddling with it again.

The `TemplateParser` still decides which exceptions to raise but the `ParsingErrorHelper` does all the noisy (awful, tedious, etc.) stuff.  I'm not committed to this separation but the parser is drowning in methods already.  

A good sign I think is that the actual source extraction and that "weird" value check both only are triggered from the helper class.


davepeck:
Thanks @ianjosephwilson -- plan to look at this soon! Will consider #164 vs #165 along the way.

davepeck:
> Thanks @ianjosephwilson -- plan to look at this soon! Will consider #164 vs #165 along the way.

And... life happened. Back in the seat; thanks for your patience.

tdom/parser.py
```diff
@@ -36,6 +37,18 @@
 )
 
 
+class ParsingError(TemplatingError):
+    pass
+
+
+class ParsingAssertionError(ParsingError):
+    pass
+
```
davepeck: Given its current uses, I think we should remove this and replace with one of (a) an actual `assert`, (b) some kind of `raise ` of a standard Python error type like `RuntimeError`. I think the `OpenTFragment` question I raised in 164 is fixed in this PR (awesome!)

tdom/parser.py
```diff
@@ -203,11 +212,219 @@ def format_starttag(self, i_index: int) -> str:
...
+    tcomponent_children: dict[TComponent, list[TNode]]
+    """ Preserved tcomponent children, copied from parser. """
+
+    sinfo_table: dict[PartPosition, TagSourceInfo]
+    """ Source info mapping, copied from parser. """
+
+    def make_mismatch_error(
+        self,
+        starttag_sinfo: OpenTagSourceInfo,
+        starttag_attrs: tuple[TAttribute, ...],
```
davepeck: seems like we can remove `starttag_attrs` now

tdom/parser.py
```diff
@@ -203,11 +212,219 @@ def format_starttag(self, i_index: int) -> str:
...
+        self, parent: OpenTag, e: ParsingError
+    ) -> None:
+        """
+        Check for cases where ambiguous slash might create a confusing error.
+
+        @NOTE: This adds exception notes to the exception but does not throw it.
+        """
+        if isinstance(
+            parent, (OpenTElement, OpenTComponent)
+        ) and self.has_ambiguous_forward_slash(parent.sinfo):
```
davepeck: now that `OpenTag` is simpler, this `isinstance()` is always `True`

tdom/parser.py
davepeck: Can we remove `format_starttag()` and `get_expression()` now?

tdom/parser.py
```diff
@@ -203,11 +212,219 @@ def format_starttag(self, i_index: int) -> str:
...
+    """
+    return ParsingErrorHelper(
+        reader=SourceReader(template=parser.get_source().template),
+        tcomponent_children=parser.tcomponent_children.copy(),
+        sinfo_table=parser.sinfo_table.copy(),
+    )
+
+
+@dataclass()
+class ParsingErrorHelper:
```
davepeck, line 211: I like this separation of concerns!

tdom/source.py
```diff
@@ -73,6 +73,16 @@ def ref_to_repr(self, ref: TemplateRef, limit: int | None = None) -> str:
         filled_template = ref.bind(self.template.interpolations)
         return template_repr(filled_template)[:limit]
 
+    def span_to_repr(self, span: TemplateSpan, limit: int | None = None) -> str:
```
davepeck: does anyone ever use the `limit`? might be nice to do it if the span is, like, a really big start tag I guess?

davepeck (commented):
Hey! Okay, a few small comments here. 

I definitely think this PR is closer to where we want to land than 164!

But I'm still considering the bigger picture; something about `run_unclosed_ambiguous_slash_checks` and friends (`get_closed_tcomps`, `tcomponent_children`, even `has_ambiguous_forward_slash`) is still feeling like... there must be a cleaner path. I just wish I saw it right now. Maybe my instinct is wrong and this problem is just annoyingly complex, not sure yet.

Also: hi! Hope things are going well on your end. :-)

tdom/parser.py
ianjosephwilson: I used `AssertionError`.

tdom/source.py
ianjosephwilson: We can bring it back later or just do it "at location".  I'm not sure if it got rebased out a while ago or just was a "might need".

ianjosephwilson:
> But I'm still considering the bigger picture; something about `run_unclosed_ambiguous_slash_checks` and friends (`get_closed_tcomps`, `tcomponent_children`, even `has_ambiguous_forward_slash`) is still feeling like... there must be a cleaner path. I just wish I saw it right now. Maybe my instinct is wrong and this problem is just annoyingly complex, not sure yet.

I'm just going to cut it out for now.  I think we have the parts necessary for the next major PR.  Once we get through that PR and get the whole thing finally wired up we can try things out including these scenarios.  I think some of the mistakes that seem insignificant might be magnified after descending through a few combinations of the processor+parser.

davepeck:
> I'm just going to cut it out for now. I think we have the parts necessary for the next major PR. Once we get through that PR and get the whole thing finally wired up we can try things out including these scenarios. 

Awesome, I think this makes for a better/narrower PR, thanks!

tdom/source.py
```diff
@@ -66,12 +66,22 @@ def values_match(self, i_index1: int, i_index2: int) -> bool:
             == self.template.interpolations[i_index2].value
         )
```
davepeck: Think we can remove this (possibly to reintroduce it in a future PR)?

tdom/parser.py
```diff
@@ -188,19 +193,79 @@ def translate_parser_span(
...
+def make_error_helper(parser: TemplateParser) -> ParsingErrorHelper:
+    """
+    Factory that creates an error helper for the parser.
+
+    @NOTE: This should only be used when an exception is already being
+    generated.
+    """
+    return ParsingErrorHelper(
+        reader=SourceReader(template=parser.get_source().template),
+        sinfo_table=parser.sinfo_table.copy(),
```
davepeck, line 206: is the `copy()` necessary?

davepeck:
@ianjosephwilson two tiny comments, but otherwise this looks really good and mergable to me -- let me know once you're comfortable and I'll commit. Thanks for all the changes and for hanging in there for my unplanned AFK-time!

davepeck:
@ianjosephwilson Also! I just invited you to join the repository directly. I turned on:

1. Requiring PRs; no more direct pushes to `main` for anyone
2. Requiring at least one PR approval before merging

The upshot: you now have the power to review and block _my_ changes, too, mwahahaha. (And if _you're_ AFK then I get to be patient and wait for a review. Seems only fair.)

Thanks for everything you're doing here!

tdom/parser.py
ianjosephwilson: We could maybe wrap it in a readonly proxy.  I guess it seems sloppy to leave it referenced with no obvious contract to not mutate it both in the helper and in the original parser if the error was caught and then the parser was manipulated.  Maybe proxy?

davepeck: Meh, I'm okay keeping it for now. We'll want to revisit someday, maybe.

davepeck (approved):
👍

tdom/source.py
ianjosephwilson: 1ad9dc74ee5e8400a591da10c4b454d732f0048f

ianjosephwilson:
@davepeck Thanks Dave.  I attempted to squash it.  Took some liberties with pruning the commit message.  Usually I just dump all the commits in the trough but I can see the benefits for a shared repo like this.

Issue #166, opened 2026-08-26
Parser error on python 3.14.7


gazpachoking:
Not exactly sure what changed here, but trying to render the following in tdom 0.1.19 works on python 3.14.6, but fails on python 3.14.7
`tdom.html(t"<button disabled={True}>x</button><button disabled={True}>y</button>")`
```
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    tdom.html(t"<button disabled={True}>x</button><button disabled={True}>y</button>")
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\username\PycharmProjects\h5connect\fsi\.venv\Lib\site-packages\tdom\processor.py", line 1047, in html
    return _default_template_processor_api.process(template, assume_ctx)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\username\PycharmProjects\h5connect\fsi\.venv\Lib\site-packages\tdom\processor.py", line 674, in process
    return self._process_template(root_template, assume_ctx)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\username\PycharmProjects\h5connect\fsi\.venv\Lib\site-packages\tdom\processor.py", line 677, in _process_template
    ttree = self.parser_api.to_ttree(template)
  File "C:\Users\username\PycharmProjects\h5connect\fsi\.venv\Lib\site-packages\tdom\processor.py", line 548, in to_ttree
    return self._to_ttree(CachableTemplate(template))
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\username\PycharmProjects\h5connect\fsi\.venv\Lib\site-packages\tdom\processor.py", line 545, in _to_ttree
    return super().to_ttree(ct.template)
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "C:\Users\username\PycharmProjects\h5connect\fsi\.venv\Lib\site-packages\tdom\processor.py", line 538, in to_ttree
    return TemplateParser.parse(template)
[...]

PR #167, opened 2026-08-27
Flush new pending data before checking if parser didn't finish.

ianjosephwilson:
I think this just gets things "working" for issue #166 but clearly it makes our dependency on the internals worse.

We use `rawdata` to check if the parser was able to complete parsing or if its waiting for more data.  We also inspect it to see if we can guess why it couldn't finish.  A common mistake is when a quote is left open like this: `t'<div x="1></div>'`.

The change in `3.14.7` adds a `_pending` buffer, `list[str]`, that lets the parser delay actual parsing until a threshold is met whereas before the parser tried to parse everytime `feed()` was called.  On `close()` this pending buffer is cleared out and parsing is done but the data is forced to `''` which I don't quite understand but we cannot detect that it wasn't able to finish.

This patch flushes the pending buffer and runs a normal parse, `self.goahead(0)` (not `self.gohead(1)`), before checking `self.rawdata` during `close()`.


davepeck:
Let's merge this now, but it feels like we should go back and do a deeper dive; sounds like we shouldn't be depending on `rawdata` to have this behavior.

ianjosephwilson:
@davepeck Have a look at #168 it tries to just cut this out for now until we can find a way forward.

davepeck:
@ianjosephwilson Agree; after looking at this a bit more, I think we need to avoid _all_ use of `rawdata`, `_pending`, `_pending_len`, etc.

